PR #12, opened 2019-10-03
Pull in label_sync from openshift

stbenjam:
Can we use this same code from openshift's test-infra (which was forked from k8s') to setup our labels? This would ensure we have consistent colors.  We'll need to run the script against the metal3-io org to sync everything. The grey labels are kind of boring to look at :-)


russellb:
I'm happy to use this.  It looks like a config file is needed before we can turn this on.  Would you like to propose some config?  If so, do you want to add it to this PR?  or want to merge this and follow up?

stbenjam:
Yea, it has the option for a cron job yaml, but it looks like OpenShift just runs it by hand (or their config is stored elsewhere) since their fork just has the default k8s config.

Something like this would create the labels on all the repos in our org.


```
bazel run //label_sync -- \
  --config $(pwd)/label_sync/labels.yaml \
  --token /path/to/github_oauth_token \
  --orgs metal3-io
```

PR #19, opened 2019-10-08
Revert "Pull in label_sync from openshift"

russellb:
Reverts metal3-io/project-infra#12


russellb:
@stbenjam I can't get this to work, so I don't want to leave it in the repo for now ... I'll file an issue to continue playing with it though

PR #21, opened 2019-10-08
Add label_sync tool from kuberenetes to sync labels

stbenjam:
This makes use of kubernetes test-infra to setup the labels on our
repos.  Copy config_example.sh and edit the values with the
configuration you desire, and then run `make` to see a *dry run* of what
it would do.

To apply the changes, run `make confirm`

### DIFF
--- a/label_sync/.gitignore
+++ b/label_sync/.gitignore
@@ -1,0 +1,4 @@
+config_*.sh
+!config_example.sh
+token--*
+test-infra/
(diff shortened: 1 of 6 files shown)


stbenjam:
https://github.com/TestOrgForLabels/myrepo/labels shows the results of running this

Happy to adjust labels.yaml as needed, I removed the things like sigs from the k8s version that I didn't think we needed.

stbenjam:
Although instead of copying the label_sync code directly, this just uses the code from kubernetes test-infra repo + a bash script + a config_XXX.sh in the style of dev-scripts. Thoughts?

russellb:
This is working!  /approve /lgtm

Issue #15, opened 2019-10-04
Broken link to test failure details


russellb:
When prow leaves a comment on a PR about a test failure, the "Details" link just links back to the PR itself instead of to something more useful.  I suspect this is missing configuration of a base URL somewhere.

Example: https://github.com/metal3-io/project-infra/pull/14

russellb:
This is now fixed - demonstrated on #25

PR #26, opened 2019-10-09
spyglass: Configure some artifact viewers.

russellb:
This should let us see the build log, start/end time, and the log of
cloning the source for the job.

Part of issue #15


russellb:
Actually, I'm just going to push this one through

PR #37, opened 2019-10-11
Setup CI as a valid label added by members and reviewers

stbenjam:


label_sync/config_example.sh
```diff
@@ -1,5 +1,5 @@
 # Which file to read the labels.yaml from
-export LABEL_SYNC_CONFIG=labels.yaml
+export LABEL_SYNC_CONFIG=../prow/config/labels.yaml
```
dhellmann: What's going on here? Is this related?

prow/config/labels.yaml
```diff
@@ -151,6 +151,12 @@ default:
       target: prs
       prowPlugin: lgtm
       addedBy: reviewers or members
+    - color: 800000
+      description: Tests a PR with Metal3 CI
+      name: CI
+      target: prs
+      prowPlugin: label
+      addedBy: reviewers or members
```
russellb: I don’t think we want this on all repos. Just the ironic image ones, and they already have the label.

label_sync/config_example.sh
stbenjam: Yes, labels.yaml was moved to config/prow to match how openshift/release has things setup. prow reads that file.

I only want reviewers or org members to be able to use `/label ci`, I'm hoping the labels.yaml is what prow uses for that. I can't find docs though.

prow/config/labels.yaml
```diff
@@ -151,6 +151,12 @@ default:
       target: prs
       prowPlugin: lgtm
       addedBy: reviewers or members
+    - color: 800000
+      description: Tests a PR with Metal3 CI
```
russellb: It’s OpenShift metal3 CI, to clarify a bit

prow/config/labels.yaml
stbenjam: I was hoping `addedBy` restricts who can add a certain label, so not anyone can trigger CI

prow/config/plugins.yaml
```diff
@@ -41,3 +41,10 @@ config_updater:
       name: config
     prow/config/plugins.yaml:
       name: plugins
+    prow/config/labels.yaml:
+      name: labels
+      key: labels.yaml
```
russellb: Why do we need this in the cluster? I haven’t deployed a label sync pod, yet anyway

stbenjam: See above about who can run CI. What restricts the `approved` and `lgtm` labels from being applied? I can't figure it out  from the prow docs, or looking at the label plugin code.

russellb: Approved and lgtm are handled by separate plugins.

stbenjam: Maybe it's not a good idea then to enable the label via prow

prow/config/labels.yaml
russellb: AFAICT there's no access control around "/label", so maybe we shouldn't enable this at all

https://github.com/kubernetes/test-infra/blob/master/prow/plugins/label/label.go

prow/config/plugins.yaml
russellb: Yeah - I just commented the same in another thread.  AFAICT, there's no access control around "/label".  Sorry to send you on a goose chase, but I guess we should leave this one as something only manually applied.

https://github.com/kubernetes/test-infra/blob/master/prow/plugins/label/label.go

stbenjam: > goose chase

Another good reason to update and get /honk!

russellb: It's no goose, but /woof

PR #39, opened 2019-10-15
Separate unit test job for CAPBM v1alpha2

maelk:
v1alpha2 unit tests require a different kubebuilder version and different utilities, hence using a different image. This comit separates the unit test job for v1alpha2 branches from master branch

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -210,6 +210,8 @@ presubmits:
   - name: unit
     always_run: true
     decorate: true
+    skip_branches:
+      - ^v1alpha2*$
     spec:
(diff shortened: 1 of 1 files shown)


prow/config/config.yaml
```diff
@@ -221,6 +223,22 @@ presubmits:
           value: "TRUE"
         image: quay.io/metal3-io/capbm-unit
```
russellb: Can you update this to use the `:master` tag?  Once we have a `v1alpha2` tag, quay.io won't automatically advance `latest`, since it doesn't know which branch to follow

PR #43, opened 2020-01-15
Add a kind/ci label

stbenjam:
During recent triage, we identified the need to mark an issue as related
to CI/testing, this adds that label.  We'll need to run the sync script
again.


russellb:
@stbenjam your account is now a metal3-io github org admin.  Can you try to run the tool?

PR #51, opened 2020-02-07
Tide : Get required jobs from branch protection

maelk:
That prevents having a PR blocking the merge pool because the
requirements are not fulfilled for merging in Github

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -43,6 +43,9 @@ tide:
     - do-not-merge/hold
     - do-not-merge/invalid-owners-file
     - do-not-merge/work-in-progress
+  context_options:
+    # Use branch protection options to define required and optional contexts
+    from-branch-protection: true
 
 deck:
   spyglass:


prow/config/config.yaml
```diff
@@ -43,6 +43,9 @@ tide:
     - do-not-merge/hold
     - do-not-merge/invalid-owners-file
     - do-not-merge/work-in-progress
+  context_options:
+    # Use branch protection options to define required and optional contexts
+    from-branch-protection: true
```
russellb: Will this make all of the prow jobs optional?

I think I've seen where we can use prow to manage branch protection, too.  Maybe we should move to that next.

maelk: I will not make other jobs optional. I checked the code, and it only adds the requirements, it does not remove the ones coming from Prow

maelk: https://github.com/kubernetes/test-infra/blob/acff7c2c6aa89cdd434b91c0bfa071537784610b/prow/config/tide.go#L541

russellb: perfect, thanks!

PR #52, opened 2020-02-10
WIP Add ironic-client repo

hardys:
Adds new repo which was discussed in https://github.com/metal3-io/ironic-image/pull/136

WIP as I've not yet pushed a PR to metal3-dev-env to consume/test this image


hardys:
https://github.com/metal3-io/metal3-dev-env/pull/215 pushed - it doesn't yet support the LOCAL_IMAGE option for this image so we'll have to figure out how to handle that if we need to test with the ironic-client image in CI (it may be reasonable to skip that in the first instance as AFAICS the openstackclient isn't used anywhere in CI atm?)

PR #54, opened 2020-02-17
Add stbenjam and maelk to OWNERS

russellb:
It's best if I'm not the only approver, and stbenjam and maelk are the
next two that have probably looked at the CI infra the most.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,7 +2,5 @@
 
 approvers:
  - russellb
-
-reviewers:
+ - maelk
  - stbenjam
- - hardys


stbenjam:
Was it intentional to remove @hardys? It's probably useful to have a few people here who can review/approve things

russellb:
> Was it intentional to remove @hardys? It's probably useful to have a few people here who can review/approve things

He was in the "reviewers" list, which doesn't do anything in practice.  I don't think he's looked much at this repo yet though.  I'm happy to add him if he starts working on the CI infra.

PR #55, opened 2020-02-24
Jenkins: use docker for e2e testing in Ubuntu only

maelk:
Otherwise use the default runtime in metal3-dev-env, podman
Podman is the only one supported in CentOS8 for now so we should
stop to enforce docker as it is not supported in CentOS8


hardys:
lgtm, thanks - it does make sense to test the natively supported container runtime on each distro I think.

PR #60, opened 2020-03-04
Add periodic jobs for marking old issues stale

stbenjam:


prow/config/config.yaml
```diff
@@ -57,6 +57,70 @@ deck:
       "clone-log.txt":
       - "buildlog"
 
+periodics:
+ - name: periodic-stale
+   interval: 1h
+   decorate: true
+   spec:
+     containers:
+     - image: gcr.io/k8s-prow/commenter:v20200304-c42b857fc
```
russellb: I think we're supposed to always use the version of prow images that matches the version of prow currently running.  See earlier in the file for examples

stbenjam: Ah I see, I didn't realize that's how it worked - updated

PR #69, opened 2020-03-27
Jenkins reducing cores to 4 for test vms

maelk:


fmuyassarov:
Seems 4 cores is enough

PR #71, opened 2020-04-07
Replaces management cluster in CI when CI host is Ubuntu

mikkosest:
Partially completes: #258


maelk:
This is not done in the right file. If you check the ubuntu CI, it ran with Minikube. The place that needs to be modified is `jenkins/scripts/files/`

mikkosest:
> this is in the wrong file

Fixed

PR #72, opened 2020-04-08
Jenkins: Use CentOS for target cluster for CentOS metal3-dev-env

maelk:
This PR introduces an override for the target image to use the same OS as for the metal3-dev-env deployment. This allows us to test CentOS deployments. It also increases the memory of the VM running the test in the case of CentOS to be able to run it.


jenkins/scripts/integration_test.sh
```diff
@@ -19,27 +19,38 @@ CI_DIR="$(dirname "$(readlink -f "${0}")")"
...
-TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-4C-16GB-100GB}"
-
+DISTRIBUTION="${DISTRIBUTION:-ubuntu}"
+if [ "${DISTRIBUTION}" == "ubuntu" ]
+then
+  IMAGE_NAME="${CI_METAL3_IMAGE}"
+  IMAGE_OS="Ubuntu"
+else
+  IMAGE_NAME="${CI_METAL3_CENTOS_IMAGE}"
+  IMAGE_OS="Centos"
```
fmuyassarov, line 29: Maybe we should change `TARGET_NODE_OS` to `"Centos"` in centos related JJB files? so that there is no need to set it explicitly here?

maelk: This is for testing for now, I don't want to break the existing tests as I am not sure that the replacement works fine.

PR #75, opened 2020-04-20
Enable kind in CI after  metal3-dev-env #293

mikkosest:


fmuyassarov:
> /test-centos-integration

I don't think editing will trigger the CI

PR #76, opened 2020-04-27
Add GITHUB_TOKEN needed  with clusterctl access to Github

mikkosest:


jenkins/jobs/integration_tests.pipeline
```diff
@@ -73,7 +73,9 @@ pipeline {
           try {
             withCredentials([usernamePassword(credentialsId: 'airshipci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')]) {
               withCredentials([sshUserPrivateKey(credentialsId: 'airshipci_city_cloud_ssh_keypair', keyFileVariable: 'AIRSHIP_CI_USER_KEY')]) {
-                sh "./jenkins/scripts/integration_test.sh"
+                withCredentials([githubToken(credentialsId: 'airshipci_city_cloud_github_token', keyFileVariable: 'GITHUB_TOKEN')]) {
```
fmuyassarov, line 76: Is the `GITHUB_TOKEN` value a string ? If yes, I think you need to use [Secret text type](https://www.jenkins.io/doc/book/using/using-credentials/) of credentials here. For that you will need: 
1. Use `string` instead of `githubToken`.  You can see example [here](https://www.jenkins.io/doc/pipeline/steps/credentials-binding/). Is there githubToken credential type in Jenkins?
2. Use `variable:` instead of `keyFileVariable:`    since the token it is not a file but a string.

jenkins/scripts/integration_test.sh
```diff
@@ -96,4 +96,4 @@ ssh \
   PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin \
   /tmp/run_integration_tests.sh "${REPO_ORG}" "${REPO_NAME}" "${REPO_BRANCH}" \
   "${UPDATED_REPO}" "${UPDATED_BRANCH}" "${CAPI_VERSION}" "${IMAGE_OS}" \
-  "${DEFAULT_HOSTS_MEMORY}" "${DISTRIBUTION}"
+  "${DEFAULT_HOSTS_MEMORY}" "${DISTRIBUTION}" "${GITHUB_TOKEN}"
```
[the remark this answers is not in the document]
mikkosest: Yes, I think it need to be added to airship jenkins secrets with admin rights, right?

jenkins/jobs/integration_tests.pipeline
mikkosest: Fixed

mikkosest: Fixed

jenkins/scripts/integration_test.sh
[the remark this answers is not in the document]
fmuyassarov: Yes correct.

fmuyassarov:
Looks good. We just only need that Jenkins credential to be added

mikkosest:
> Looks good. We just only need that Jenkins credential to be added

  Do you know, who to ask?

fmuyassarov:
> > Looks good. We just only need that Jenkins credential to be added
> 
> Do you know, who to ask, who has Jenkins rights nowadays?

I know @kanwar-saad and also @maelk I guess ?

jenkins/jobs/integration_tests.pipeline
```diff
@@ -73,7 +73,9 @@ pipeline {
           try {
             withCredentials([usernamePassword(credentialsId: 'airshipci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')]) {
               withCredentials([sshUserPrivateKey(credentialsId: 'airshipci_city_cloud_ssh_keypair', keyFileVariable: 'AIRSHIP_CI_USER_KEY')]) {
-                sh "./jenkins/scripts/integration_test.sh"
+                withCredentials([string(credentialsId: 'airshipci_city_cloud_github_token', variable: 'GITHUB_TOKEN')]) {
```
maelk, line 76: The token you need already exists : nordix-airship-ci-github-prod-token

mikkosest: Fixed

mikkosest:
> /lgtm
> So, I understood that the token as not a String text type but Username/password type right?

Yes, it complained about issue when using string

fmuyassarov:
> I am actually having second thoughts about exposing the token like this. We should probably create a new token indeed, without any rights, just to authenticate, so that if it is leaked through metal3-dev-env, the impact would be non-existent

That make sense to make to have a separate token only for this purpose.

PR #79, opened 2020-04-30
prow: Add branch protection management

russellb:
Branch protection makes it so prow (tide) will not merge a PR unless a
set of requirement CI jobs passes.  We currently manage this manually
in github to ensure the functional CI jobs from jenkins passes.  This
PR moves the configuration of what jobs block PRs into this repo and
to be managed by prow.

All jobs that run by default (always_run: true) will now block by
default.  Jobs that run outside of prow (like jenkins) must be
explicitly listed in configuration.

We weren't actually blocking on any of the prow CI jobs before this,
which is one major improvement.  This also lets us manage another
aspect of our github org configuration through this repo.

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -48,6 +48,49 @@ tide:
     # Use branch protection options to define required and optional contexts
     from-branch-protection: true
 
+#
+# Use prow to manage the configuration of github branch protection.
+# This is what controls which github status checks (or CI jobs) must
+# be passing for a PR to merge.
+#
+# Docs: https://github.com/kubernetes/test-infra/tree/master/prow/cmd/branchprotector
+#
+branch-protection:
+  orgs:
+    metal3-io:
+      # Require "always_run: true" jobs to pass before merging.
+      # To turn this off for a given job, set "optional: true"
+      # in the job definition.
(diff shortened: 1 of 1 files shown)


prow/config/config.yaml
```diff
@@ -48,6 +48,36 @@ tide:
...
+      repos:
+        baremetal-operator:
+          # Use this to specify that a status coming from outside of prow is
+          # required.  We use this to require functional jobs running in
+          # jenkins are required, for example.
+          required_status_checks:
+            contexts: ["test-integration"]
+        cluster-api-provider-metal3:
+          required_status_checks:
+            contexts: ["test-v1a4-integration"]
```
maelk, line 67: Can we specify per branch ? for the release-0.3 branch it would be test-integration. In this current shape, it might prevent us from merging any PR on that branch.

russellb: Thanks for catching this!  I'll figure how how to do branch settings and post a revision.

russellb: done.  Take a look and let me know if I've set the right policy.  I have CAPM3 set for `test-integration` except on master it's `test-v1a4-integration`.  I can tweak further if needed.

PR #84, opened 2020-05-06
Temporary patch for ironic in CI

mikkosest:


mikkosest:
test-centos-integration

fmuyassarov:
> @fmuyassarov do we still need this patch or we can close it maybe?

@mikkosest Can we close it if you think we don't need this patch? Otherwise it will be marked as lifecycle stale after 90 days of inactivity.

PR #88, opened 2020-05-09
jenkins: Add fetching container logs from metal3-dev-env

maelk:
This PR modifies the pipeline and the scripts running the metal3-dev-env integration tests to fetch the logs from all containers at the end of the run, they will be saved as artifacts for that run

### DIFF
--- a/jenkins/jobs/integration_tests.pipeline
+++ b/jenkins/jobs/integration_tests.pipeline
@@ -38,6 +38,8 @@ pipeline {
     OS_IDENTITY_API_VERSION=3
     TEST_EXECUTER_VM_NAME = "${VM_NAME}"
     BUILD_URL = "${env.BUILD_URL}"
+    BUILD_NUMBER = "${env.BUILD_NUMBER}"
+    JOB_NAME = "${env.JOB_BASE_NAME}"
     PR_ID = "${ghprbPullId}"
     DISTRIBUTION = "${DISTRIBUTION}"
     CAPI_VERSION = "${CAPI_VERSION}"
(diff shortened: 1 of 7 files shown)


jenkins/scripts/files/run_fetch_logs.sh
```diff
@@ -0,0 +1,37 @@
...
+mkdir -p logs/docker
+DOCKER_CONTAINERS="$(docker ps --format "{{json .}}" | jq -r '.Names')"
+for DOCKER_CONTAINER in $DOCKER_CONTAINERS
+do
+  mkdir -p "logs/docker/${DOCKER_CONTAINER}"
+  docker logs "$DOCKER_CONTAINER" > "logs/docker/${DOCKER_CONTAINER}/stdout.log" \
+  2> "logs/docker/${DOCKER_CONTAINER}/stderr.log"
+done
+
+tar -cvzf "$LOGS_TARBALL" logs/*
```
dhellmann, line 37: Is there a way to name the "logs" directory uniquely based on the job so that when the tarball is unpacked it creates a unique directory locally?

maelk: good point! I'll add that.

jenkins/jobs/integration_tests.pipeline
```diff
@@ -38,6 +38,8 @@ pipeline {
     OS_IDENTITY_API_VERSION=3
     TEST_EXECUTER_VM_NAME = "${VM_NAME}"
     BUILD_URL = "${env.BUILD_URL}"
+    BUILD_NUMBER = "${env.BUILD_NUMBER}"
+    JOB_NAME = "${env.JOB_BASE_NAME}"
```
fmuyassarov, line 42: you could probably use BUILD_TAG which covers both BUILD_NUMBER and JOB_NAME.

jenkins/jobs/integration_tests.pipeline
```diff
@@ -94,11 +96,26 @@ pipeline {
               echo "${err}"
               currentBuild.result = 'FAILURE'
           }
+          finally {
```
fmuyassarov, line 99: what if job fails, will we still be able to run the final script to fetch the logs?

fmuyassarov:
Apart from above comments, looks good.

jenkins/jobs/integration_tests.pipeline
maelk: yes, that's why there is a finally

jenkins/jobs/integration_tests.pipeline
maelk: done, thank you

PR #94, opened 2020-05-14
jenkins: increase node memory to 8GB for Ubuntu deployments

maelk:
With the new IPA (current master), 8GB is required to deploy Ubuntu


jenkins/scripts/integration_test.sh
```diff
@@ -36,12 +36,7 @@ UPDATED_REPO="${UPDATED_REPO:-https://github.com/${REPO_ORG}/${REPO_NAME}.git}"
 UPDATED_BRANCH="${UPDATED_BRANCH:-master}"
 CAPI_VERSION="${CAPI_VERSION:-v1alpha3}"
 
-if [ "${IMAGE_OS}" == "Ubuntu" ]
-then
-  DEFAULT_HOSTS_MEMORY="${DEFAULT_HOSTS_MEMORY:-6144}"
-else
-  DEFAULT_HOSTS_MEMORY="${DEFAULT_HOSTS_MEMORY:-8192}"
-fi
+DEFAULT_HOSTS_MEMORY="${DEFAULT_HOSTS_MEMORY:-8192}"
```
juliakreger, line 39: I guess we're not raw streaming :\

maelk: No, that is not in place, but I have been looking into that, and we could just convert the image after downloading them. we'd need adjustments on the metal3-dev-env side. that's the next step.

fmuyassarov, line 39: I think you need to change also `TARGET_NODE_MEMORY` in capi_bm_v1aX_integration_tests_ubuntu.yaml in gerrit, which is defaulted to 4096.

fmuyassarov: Because the initial value is inherent from from here `TARGET_NODE_MEMORY` in [here](https://github.com/metal3-io/project-infra/blob/ad99716fcf645498ff133880f87b9978dbaa89c5/jenkins/jobs/integration_tests.pipeline#L47).

maelk: hum, maybe we just override it for now, as it is hopefully a temporary workaround until we have raw image streaming in place. Otherwise, the TARGET_NODE_MEMORY parameter does not make sense anyways, since we just base the size on the target distribution

maelk: very good catch, thank you

PR #97, opened 2020-05-16
Add some details on who can run each prow command.

russellb:


maelk:
lgtm. Should the owners file part be a bit expanded to cover the per directory cases ?

russellb:
> lgtm. Should the owners file part be a bit expanded to cover the per directory cases ?

updated with a link to the OWNERS file docs

PR #105, opened 2020-05-21
Address issues encountered with prow upgrade

russellb:
After merging and applying #103 to our CI cluster, two prow components failed to start. One needed a new argument and another needed a tweak to configuration.


russellb:
I've pushed a revert of the upgrade into this PR.

Based on the remaining errors I was seeing from `plank`, I think I need to set up https://github.com/kubernetes/test-infra/tree/master/ghproxy#ghproxy before trying again.

PR #108, opened 2020-06-01
Jenkins, fetch m3-dev-env console logs during log gathering

maelk:
Following https://github.com/metal3-io/metal3-dev-env/pull/330 we can fetch the vm console logs as part of the log fetching process in the metal3-dev-env tests.


stbenjam:
Nice!

/lgtm

PR #114, opened 2020-06-07
Jenkins: fix wrong condition for master runs

maelk:
There was a mistake in the condition determining if it is a master run


maelk:
@fmuyassarov I now see what you meant with your comment on the previous PR.
/cc @fmuyassarov 
/cc @furkatgofurov7 
/cc @kashifest

PR #115, opened 2020-06-12
Increase timeout for feature test

kashifest:
Increase timeout for feature tests to 10 hours since now the upgrade tests will also be running.

### DIFF
--- a/jenkins/jobs/integration_tests.pipeline
+++ b/jenkins/jobs/integration_tests.pipeline
@@ -7,7 +7,7 @@ def TIMEOUT = 5400
 
 if (env.TESTS_FOR) {
   if ("${TESTS_FOR}" == "feature_tests") {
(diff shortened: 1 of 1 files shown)


kashifest:
> 10 hrs sound very long. We should maybe look into a different directions to get all those tests running. We should maybe spend some time improving the test setup time (for example not using images but volumes, cloned for each new machine, so that we can build in all the images and slice down the 01 script execution time)

I agree

fmuyassarov:
> 10 hrs sound very long. We should maybe look into a different directions to get all those tests running. We should maybe spend some time improving the test setup time (for example not using images but volumes, cloned for each new machine, so that we can build in all the images and slice down the 01 script execution time)

I definitely agree that 10 hours is too long. That can be a task as part of optimization.

PR #117, opened 2020-06-18
Remove unnecessary vars Update default memory

kashifest:
Remove `IMAGE_OS` variable since it is not needed anymore.

The following [PR](https://gerrit.nordix.org/c/infra/cicd/+/5091. 
) removes populating the `TARGET_NODE_OS` 

Also updating the `DEFAULT_HOSTS_MEMORY` to 8192.


jenkins/jobs/integration_tests.pipeline
```diff
@@ -51,7 +51,6 @@ pipeline {
     PR_ID = "${ghprbPullId}"
     DISTRIBUTION = "${DISTRIBUTION}"
     CAPI_VERSION = "${CAPI_VERSION}"
-    IMAGE_OS = "${TARGET_NODE_OS}"
```
maelk, line 54: This variable is used in https://github.com/metal3-io/project-infra/pull/117/files#diff-df6d7fae6b1eb79d8eff8d9f7869a586R101

fmuyassarov: I'm not sure if this is a correct link, but even if it is removed here IMAGE_OS will be declared in
[integration_test.sh](https://github.com/metal3-io/project-infra/blob/f8c482a682194795aa172371c35b8ec66b8d7a3a/jenkins/scripts/integration_test.sh#L26).

fmuyassarov: We can have two separate jobs that will be running for example Ubuntu for target nodes when CentOS is used for the ephemeral cluster VM and vice versa.

kashifest: The link is incorrect @maelk

maelk: https://github.com/metal3-io/project-infra/blob/master/jenkins/scripts/integration_test.sh#L101

maelk: so it should either be removed in both places of that file, and rely on the override in integration_test.sh, or remove the override. I would rather remove the override.

kashifest: Changed. Please review again

fmuyassarov: Wouldn’t it be better if we just drop the `TARGET_NODE_OS` variable from JJB files and declare `IMAGE_OS` here in project-infra instead of converting `TARGET_NODE_OS` to `IMAGE_OS` ? The only reason I see to keep `TARGET_NODE_OS` would be when for example  `DISTRIBUTION=”Ubuntu”` and we want test both ubuntu and centos for target nodes, which is something we  aren’t doing.

maelk: I would keep the TARGET_NODE_OS to at least have the possibility to test with cross distributions (centos on Ubuntu and vice versa). Even if we don't do it, it might come handy at some point to be able to do it.

jenkins/scripts/integration_test.sh
```diff
@@ -23,10 +23,10 @@ DISTRIBUTION="${DISTRIBUTION:-ubuntu}"
 if [ "${DISTRIBUTION}" == "ubuntu" ]
 then
   IMAGE_NAME="${CI_METAL3_IMAGE}"
-  IMAGE_OS="Ubuntu"
+  IMAGE_OS="${IMAGE_OS:-Ubuntu}"
```
maelk, line 26: you could just remove this line

jenkins/scripts/integration_test.sh
```diff
@@ -23,10 +23,10 @@ DISTRIBUTION="${DISTRIBUTION:-ubuntu}"
...
 else
   IMAGE_NAME="${CI_METAL3_CENTOS_IMAGE}"
-  IMAGE_OS="Centos"
+  IMAGE_OS="${IMAGE_OS:-Centos}"
```
maelk, line 29: same here

PR #119, opened 2020-06-22
Jenkins: reduce VM RAM size to 10 / 16 GB

maelk:
Ubuntu : from 16 to 10 GB
CentOS : from 32 to 16 GB


jenkins/scripts/integration_test.sh
```diff
@@ -49,9 +49,9 @@ fi
 
 if [ "${IMAGE_OS}" == "Ubuntu" ]
 then
-  TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-4C-16GB-200GB}"
+  TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-4C-8GB-200GB}"
```
fmuyassarov, line 52: I'm wondering if we can reduce it to 16GB in line 47 as well

maelk: I didn't want to touch something that I have not been working with. In theory, I would say yes, but I'll leave it up to you.

fmuyassarov: Yes, sure. I'll check that.

fmuyassarov:
Centos job is failing because of a bug in metal3-dev-env, not related to this PR. That issue should be fixed in [#350](https://github.com/metal3-io/metal3-dev-env/pull/350)

PR #120, opened 2020-06-22
Jenkins: fix name issue in cleanup script

maelk:
The script was not removing the random string after the date


maelk:
test-integration does not test this script

PR #124, opened 2020-07-07
Upgrade golang containers to 1.15.3

honza:


maelk:
I think this should be matched with updates of the hack scripts in all repos too. Otherwise we'll have differences there. 

/approve

furkatgofurov7:
@honza I have opened corresponding PRs' in other repos and upgraded go to 1.15.3 which is the latest currently, do you think we can do it the same for this patch as well to match the version?

honza:
@furkatgofurov7 Done, thanks for the nudge

furkatgofurov7:
@honza welcome! Can you please update PR title also, it has old version tag. Thanks!

PR #126, opened 2020-07-17
✨Add using volume instead of image to integration tests

furkatgofurov7:
This patch takes the volumes created in [airship-dev-tools/pull/210](https://github.com/Nordix/airship-dev-tools/pull/210)
in use, by cloning the volume and starting the VM in CI with the cloned base volumes

Related to this patch:

-  [airship-dev-tools/pull/219](https://github.com/Nordix/airship-dev-tools/pull/219)

### DIFF
--- a/jenkins/scripts/integration_clean.sh
+++ b/jenkins/scripts/integration_clean.sh
@@ -28,11 +28,22 @@ echo "Cleaning old resources"
 
 for VM_NAME in $VM_LIST
 do
-  # Delete executer vm
+  # Delete executer VM
   echo "Deleting executer VM ${VM_NAME}."
   openstack server delete "${VM_NAME}"
 done
 
+VOLUME_LIST=$(openstack volume list -f json | jq -r '.[] | select(.Name |
+  startswith("ci-test-volume-")) | select((.Name | ltrimstr("ci-test-volume-") |
+  split("-") | .[0] | strptime("%Y%m%d%H%M%S") | mktime) < (now - 21600))
+  | .ID ')
+
+for VOLUME_NAME in $VOLUME_LIST
+do
+  # Delete executer volume
+  echo "Deleting executer volume ${VOLUME_NAME}."
+  openstack volume delete --force "${VOLUME_NAME}"
+done
 
 PORT_LIST=$(openstack port list -f json | jq -r '.[] | select(.Name |
   startswith("ci-test-vm-")) | select((.Name | ltrimstr("ci-test-vm-") |
@@ -41,7 +52,7 @@ PORT_LIST=$(openstack port list -f json | jq -r '.[] | select(.Name |
 
 for PORT_NAME in $PORT_LIST
 do
-  # Delete executer vm
+  # Delete executer VM port
   echo "Deleting executer VM port ${PORT_NAME}."
   openstack port delete "${PORT_NAME}"
 done
--- a/jenkins/scripts/integration_delete.sh
+++ b/jenkins/scripts/integration_delete.sh
@@ -17,8 +17,26 @@ source "${CI_DIR}/utils.sh"
 
 TEST_EXECUTER_PORT_NAME="${TEST_EXECUTER_PORT_NAME:-${TEST_EXECUTER_VM_NAME}-int-port}"
 
-# Delete executer vm
+# Delete executer VM
 echo "Deleting executer VM."
 openstack server delete "${TEST_EXECUTER_VM_NAME}"
 
+echo "Waiting until volume status is available, to proceed with proper volume deletion."
+until openstack volume show "${TEST_EXECUTER_VM_NAME}" -f json \
+  | jq .status | grep "available"
+do
+  sleep 10
+  if [[ "$(openstack volume show "${TEST_EXECUTER_VM_NAME}" -f json \
+    | jq .status)" == "error" ]];
+  then
+    exit 1
+  fi
+done
+
+# Delete executer volume
+echo "Deleting executer volume."
+openstack volume delete "${TEST_EXECUTER_VM_NAME}"
+
+# Delete executer VM port
+echo "Deleting executer VM port."
 openstack port delete "${TEST_EXECUTER_PORT_NAME}"
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -23,8 +23,10 @@ DISTRIBUTION="${DISTRIBUTION:-ubuntu}"
 if [ "${DISTRIBUTION}" == "ubuntu" ]
 then
   IMAGE_NAME="${CI_METAL3_IMAGE}"
+  BASE_VOLUME_NAME="metal3-ubuntu"
 else
   IMAGE_NAME="${CI_METAL3_CENTOS_IMAGE}"
+  BASE_VOLUME_NAME="metal3-centos"
 fi
 
 REPO_ORG="${REPO_ORG:-metal3-io}"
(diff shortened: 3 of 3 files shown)


jenkins/scripts/integration_test.sh
```diff
@@ -38,7 +38,17 @@ TESTS_FOR="${TESTS_FOR:-integration_test}"
 
 DEFAULT_HOSTS_MEMORY="${DEFAULT_HOSTS_MEMORY:-4096}"
 
+if [ "${DISTRIBUTION}" == "ubuntu" ]
+then
+  BASE_VOLUME_NAME="metal3-ubuntu"
+  TEST_EXECUTER_VM_FLAVOR="2C-4GB-50GB"
```
fmuyassarov, line 44: We already have a variable declaration for flavor [here](https://github.com/metal3-io/project-infra/blob/a5671bca76e832e4225f8e6a4f43345a19d91300/jenkins/scripts/integration_test.sh#L50). Is there a need for creating another variable for flavor?

fmuyassarov: And I don't think that 50 GB would be enough for successfully run.

furkatgofurov7: > We already have a variable declaration for flavor [here](https://github.com/metal3-io/project-infra/blob/a5671bca76e832e4225f8e6a4f43345a19d91300/jenkins/scripts/integration_test.sh#L50). Is there a need for creating another variable for flavor?



> And I don't think that 50 GB would be enough for successfully run.

Yes, but ideally, should not [this](https://github.com/metal3-io/project-infra/blob/a5671bca76e832e4225f8e6a4f43345a19d91300/jenkins/scripts/integration_test.sh#L50) go to `else` of this [condition](https://github.com/metal3-io/project-infra/blob/a5671bca76e832e4225f8e6a4f43345a19d91300/jenkins/scripts/integration_test.sh#L45)? 
What do you think of memory required to be able to successfully run it?

fmuyassarov: Currently, we are using the same flavor for both Ubuntu & Centos (4C-16GB-100GB). And probably you should still use the same. And once all the installations are moved to a single place, then we can test it again to know if we we can reduce the resource consumption.

fmuyassarov: > > We already have a variable declaration for flavor [here](https://github.com/metal3-io/project-infra/blob/a5671bca76e832e4225f8e6a4f43345a19d91300/jenkins/scripts/integration_test.sh#L50). Is there a need for creating another variable for flavor?
> 
> > And I don't think that 50 GB would be enough for successfully run.
> 
> Yes, but ideally, should not [this](https://github.com/metal3-io/project-infra/blob/a5671bca76e832e4225f8e6a4f43345a19d91300/jenkins/scripts/integration_test.sh#L50) go to `else` of this [condition](https://github.com/metal3-io/project-infra/blob/a5671bca76e832e4225f8e6a4f43345a19d91300/jenkins/scripts/integration_test.sh#L45)?
It can. However, L50 first checks if `TEST_EXECUTER_FLAVOR` var initialized, if yes that is done for departure tests, if not then it will assume that the run is for integration test and assign the default value.

furkatgofurov7: Thanks!

jenkins/scripts/integration_test.sh
```diff
@@ -47,6 +48,13 @@ then
   TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-300GB}"
 fi
 
+if [ "${DISTRIBUTION}" == "ubuntu" ]
+then
+  BASE_VOLUME_NAME="metal3-ubuntu"
+else
+  BASE_VOLUME_NAME="metal3-centos"
+fi
+
```
fmuyassarov, line 57: You could use already existing conditional check in L23.

jenkins/scripts/integration_test.sh
```diff
@@ -57,6 +60,22 @@ EXT_PORT_ID="$(openstack port create -f json \
...
+# Create new executer vm from copy of the test executer volume
+openstack server create \
+  --volume "${TEST_EXECUTER_VOLUME_ID}" \
+  --flavor "${TEST_EXECUTER_FLAVOR}" \
+  --port "${EXT_PORT_ID}" \
+  "${TEST_EXECUTER_VM_NAME}"
+
 # Create new executer vm
 openstack server create -f json \
```
fmuyassarov, line 80: You don't need these lines, since you create a VM in L73.

furkatgofurov7: Yeah agree, but did not we agree to do a cleanup of additional code after these changes are in place, in order to have a backwards compatibility?

fmuyassarov: What I meant previously is that this patch should go after we start building volumes with the image building CI. Once volumes started to be created in CityCloud by the image building CI,  we trigger the integration CI on this PR to check if everything goes like clockwork.

furkatgofurov7: Okay, removed.

jenkins/scripts/integration_delete.sh
```diff
@@ -17,8 +17,14 @@ source "${CI_DIR}/utils.sh"
...
 
-# Delete executer vm
+# Delete executer VM
 echo "Deleting executer VM."
 openstack server delete "${TEST_EXECUTER_VM_NAME}"
 
-openstack port delete "${TEST_EXECUTER_PORT_NAME}"
+# Delete executer volume
+echo "Deleting executer volume."
+openstack volume delete "${TEST_EXECUTER_VOLUME_NAME}"
```
maelk, line 26: this is not defined here

furkatgofurov7: Fixed

jenkins/scripts/integration_delete.sh
```diff
@@ -15,10 +15,18 @@ CI_DIR="$(dirname "$(readlink -f "${0}")")"
 # shellcheck disable=SC1090
 source "${CI_DIR}/utils.sh"
 
+VM_TIMELABEL="${VM_TIMELABEL:-$(date '+%Y%m%d%H%M%S')}"
+TEST_EXECUTER_VOLUME_NAME="${TEST_EXECUTER_VOLUME_NAME:-ci-test-volume-${VM_TIMELABEL}}"
```
maelk, line 19: This is problematic, because the VM_TIMELABEL variable is actually not defined in the pipeline. so your name would mismatch. You should instead use the same approach as for the vm name (if not directly using the vm name instead of a volume name) : https://github.com/metal3-io/project-infra/blob/master/jenkins/jobs/integration_tests.pipeline#L49

jenkins/scripts/integration_test.sh
```diff
@@ -39,6 +41,7 @@ TESTS_FOR="${TESTS_FOR:-integration_test}"
 DEFAULT_HOSTS_MEMORY="${DEFAULT_HOSTS_MEMORY:-4096}"
 
 VM_TIMELABEL="${VM_TIMELABEL:-$(date '+%Y%m%d%H%M%S')}"
+TEST_EXECUTER_VOLUME_NAME="${TEST_EXECUTER_VOLUME_NAME:-ci-test-volume-${VM_TIMELABEL}}"
```
maelk, line 44: that variable should be defined in the pipeline to not default to unpredictable name : https://github.com/metal3-io/project-infra/blob/master/jenkins/jobs/integration_tests.pipeline#L49. you can also directly use the vm name, instead of having another name for the volume

jenkins/scripts/integration_test.sh
```diff
@@ -57,12 +60,21 @@ EXT_PORT_ID="$(openstack port create -f json \
...
-openstack server create -f json \
-  --image "${IMAGE_NAME}" \
+# Get the base volume
+BASE_VOLUME_NAME_ID="$(openstack volume show -f json \
+  "${BASE_VOLUME_NAME}" | jq -r '.id')"
+
+# Create copy of base volume
+TEST_EXECUTER_VOLUME_ID="$(openstack volume create -f json \
+  --source "${BASE_VOLUME_NAME_ID}" \
+  "${TEST_EXECUTER_VOLUME_NAME}" | jq -r '.id')"
```
maelk, line 70: you should specify a size here probably, since 50GB won't be enough

maelk: Also since you use a volume, you probably do not need the volume part of https://github.com/metal3-io/project-infra/pull/126/files#diff-df6d7fae6b1eb79d8eff8d9f7869a586R50 and https://github.com/metal3-io/project-infra/pull/126/files#diff-df6d7fae6b1eb79d8eff8d9f7869a586R53

jenkins/scripts/integration_delete.sh
furkatgofurov7: I will directly use the vm name instead of declaring and using another name for a volume

jenkins/scripts/integration_test.sh
furkatgofurov7: Directly using the vm name

maelk:
LGTM, let's wait to have the source volumes in place before running the tests

jenkins/scripts/integration_test.sh
furkatgofurov7: > you should specify a size here probably, since 50GB won't be enough

I have specified a size of 200GB

> Also since you use a volume, you probably do not need the volume part of https://github.com/metal3-io/project-infra/pull/126/files#diff-df6d7fae6b1eb79d8eff8d9f7869a586R50 and https://github.com/metal3-io/project-infra/pull/126/files#diff-df6d7fae6b1eb79d8eff8d9f7869a586R53

Sorry, but there are no volume parts in the links you are referencing to?

maelk: it's because you need to expand the file. Anyways, I meant this :
```
if [ "${TESTS_FOR}" == "feature_tests" ]
then
  TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-300GB}"
fi
TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-4C-16GB-200GB}"
```

furkatgofurov7: Done.

furkatgofurov7:
/hold 
until [airship-dev-tools#210](https://github.com/Nordix/airship-dev-tools/pull/210) is merged to properly test this one

maelk:
@furkatgofurov7 you'll need to investigate the cause of the failure.

furkatgofurov7:
/hold
testing locally with centos to find out problem for failing

jenkins/scripts/integration_test.sh
```diff
@@ -47,30 +49,55 @@ if [ "${TESTS_FOR}" == "feature_tests" ]
...
+  "${BASE_VOLUME_NAME}" | jq -r '.id')"
+
+# Create test executer volume from copy of base volume
+echo "Creating a test executer volume from copy of base volume."
+TEST_EXECUTER_VOLUME_ID="$(openstack volume create -f json \
+  --source "${BASE_VOLUME_NAME_ID}" \
+  --size 200 \
+  "${TEST_EXECUTER_VM_NAME}" | jq -r '.id')"
+
+if [ "${BASE_VOLUME_NAME}" == "metal3-centos" ]
```
furkatgofurov7, line 73: @maelk would it be a good idea to try like this?

maelk: no, this would run on the Jenkins node, not on the test vm. I was talking about the airship-dev-tools volume building script

furkatgofurov7: sorry, misunderstood, please check [this](https://github.com/Nordix/airship-dev-tools/pull/217)

Jaakko-Os:
looks good to me

jenkins/scripts/integration_delete.sh
```diff
@@ -17,8 +17,21 @@ source "${CI_DIR}/utils.sh"
...
 echo "Deleting executer VM."
 openstack server delete "${TEST_EXECUTER_VM_NAME}"
 
-openstack port delete "${TEST_EXECUTER_PORT_NAME}"
+echo "Waiting until volume status is available, to proceed with proper volume deletion."
+until openstack volume show "${TEST_EXECUTER_VM_NAME}" -f json \
+  | jq .status | grep "available"
+do
+  sleep 10
+done
```
maelk, line 29: you should maybe add an exit condition here. If the status is error, it will probably not get out of that on its own, and you should proceed with deletion.

jenkins/scripts/integration_clean.sh
```diff
@@ -28,11 +28,22 @@ echo "Cleaning old resources"
...
+VOLUME_LIST=$(openstack volume list -f json | jq -r '.[] | select(.Name |
+  startswith("ci-test-volume-")) | select((.Name | ltrimstr("ci-test-volume-") |
+  split("-") | .[0] | strptime("%Y%m%d%H%M%S") | mktime) < (now - 21600))
+  | .ID ')
+
+for VOLUME_NAME in $VOLUME_LIST
+do
+  # Delete executer volume
+  echo "Deleting executer volume ${VOLUME_NAME}."
+  openstack volume delete "${VOLUME_NAME}"
```
maelk, line 45: using --force might make sense here, in case the volume is in an unexpected state, maybe ?

furkatgofurov7: @maelk Is it only specific to volume or --force is needed also for server,port deletions as well?

maelk: Well, it is not specific to volumes, but we have seen volumes reach some problematic states,while the others are usually not going into states where it cannot be deleted.

furkatgofurov7: Yes, true, added. Thanks!

Issue #128, opened 2020-07-23
Gather cluster resources as well as logs in CI


stbenjam:
It would be useful to collect BareMetalHost, Machine, and Node YAML as part of the artifacts created in CI.

PR #129, opened 2020-07-24
Jenkins: fetch manifest at the of CI run

fmuyassarov:
Fixes https://github.com/metal3-io/project-infra/issues/128

### DIFF
--- a/jenkins/scripts/files/run_fetch_logs.sh
+++ b/jenkins/scripts/files/run_fetch_logs.sh
@@ -14,6 +14,10 @@ else
 fi
 mkdir -p ${LOGS_DIR}
 
+# Fetch cluster manifests
+mkdir -p "${LOGS_DIR}/manifests"
+cp -r /tmp/manifests/* "${LOGS_DIR}/manifests"
+
 NAMESPACES="$(kubectl get namespace -o json | jq -r '.items[].metadata.name')"
 mkdir -p "${LOGS_DIR}/k8s"
 for NAMESPACE in $NAMESPACES


jenkins/scripts/files/run_fetch_logs.sh
```diff
@@ -14,6 +14,24 @@ else
 fi
 mkdir -p ${LOGS_DIR}
 
+# Fetch manifests
+declare -a resource=("bmh" \
+"machine" \
+"m3m" \
+"cluster" \
+"m3cluster" \
+"node")
```
maelk, line 23: that would not be run on the target cluster, so you would get the node of the management cluster (kind or minikube)

maelk: I would also add all CAPI objects, such as KCP, Machine Deployment, Machine set, Kubeadmconfig (inc. all templates).

jenkins/scripts/files/run_fetch_logs.sh
```diff
@@ -14,6 +14,24 @@ else
...
+
+for object in "${resource[@]}"; do
+    mkdir -p "${LOGS_DIR}/${object}"
```
maelk, line 26: can you make it under  another directory, like "objects"  or anything similar ? to avoid adding many folders in the root of the archive

jenkins/scripts/files/run_fetch_logs.sh
fmuyassarov: > I would also add all CAPI objects, such as KCP, Machine Deployment, Machine set, Kubeadmconfig (inc. all templates).

Done.

fmuyassarov: > that would not be run on the target cluster, so you would get the node of the management cluster (kind or minikube)

Do you think we exclude that? Or fetch it from the target cluster? I'm not sure if we really need nodes.

jenkins/scripts/files/run_fetch_logs.sh
fmuyassarov: I created a folder `manifest`, and adding everything under it. Let me if you think it is not good idea.

jenkins/scripts/files/run_fetch_logs.sh
maelk: you can start without it, and we can add it later on.

fmuyassarov:
> This is not run at the proper time. At that point the deprovisioning has already happened so those objects do not exist anymore

Thanks a lot. I missed that. 
Created a [#392](https://github.com/metal3-io/metal3-dev-env/pull/392)

fmuyassarov:
/hold cancel
metal3-io/metal3-dev-env#392 is merged now.

PR #134, opened 2020-07-30
Update prow config for CAPM3 jobs

maelk:
After the CAPM3 v1a4 release, we have changed or main jobs to v1a4. We need to adapt the prow config to reflect the change in CAPM3


maelk:
Considering that both other reviewers are not reachable this week, I'll proceed with a manual merge. @fmuyassarov looks good to you ?

PR #139, opened 2020-08-21
CI: Fix fetching qemu logs in Centos

Jaakko-Os:
Fixes: https://github.com/metal3-io/metal3-dev-env/issues/438

### DIFF
--- a/jenkins/scripts/files/run_fetch_logs.sh
+++ b/jenkins/scripts/files/run_fetch_logs.sh
@@ -48,7 +48,7 @@ do
 done
 
 mkdir -p "${LOGS_DIR}/qemu"
-sudo cp -r /var/log/libvirt/qemu/* "${LOGS_DIR}/qemu/"
+sudo sh -c "cp -r /var/log/libvirt/qemu/* "${LOGS_DIR}/qemu/""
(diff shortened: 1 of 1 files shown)


smoshiur1237:
Looks good to me

fmuyassarov (approved):

jenkins/scripts/files/run_fetch_logs.sh
```diff
@@ -48,7 +48,7 @@ do
 done
 
 mkdir -p "${LOGS_DIR}/qemu"
-sudo cp -r /var/log/libvirt/qemu/* "${LOGS_DIR}/qemu/"
+sudo sh -c "cp -r /var/log/libvirt/qemu/* "${LOGS_DIR}/qemu/""
```
maelk, line 51: There is an issue with the quotes here. you should escape the inner ones, otherwise ${LOGS_DIR/qemu/ is not quoted and you have a useless double quote marks at the end. @Jaakko-Os Can you please fix this in a new PR ?

Jaakko-Os: https://github.com/metal3-io/project-infra/pull/140

PR #142, opened 2020-09-11
CI: reduce feature_test timeout

Jaakko-Os:
Based on experience the timeout value used by
pivoting, remediation and upgrade can be reduced
from 36000 sec (10 hours) to 21600 sec (6 hours)


namnx228 (approved):

furkatgofurov7 (commented):
lgtm

This should be fine since we have separated feature and upgrade tests

fmuyassarov (approved):

Jaakko-Os:
/hold
until CI slowness is cleared

PR #148, opened 2020-09-23
prow: run gomod.sh as a presubmit

honza:
Depends on https://github.com/metal3-io/baremetal-operator/pull/648

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -259,6 +259,20 @@ presubmits:
           value: "TRUE"
         image: registry.hub.docker.com/library/golang:1.13.7
         imagePullPolicy: Always
+  - name: gomod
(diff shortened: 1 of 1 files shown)


prow/config/config.yaml
```diff
@@ -259,6 +259,20 @@ presubmits:
           value: "TRUE"
         image: registry.hub.docker.com/library/golang:1.13.7
         imagePullPolicy: Always
+  - name: gomod
```
dhellmann, line 262: It looks like this would only apply to the BMO, right? Could you update the commit message to indicate that?

honza: Done

russellb:
/approve

someone else can lgtm when you're ready for it to merge

PR #155, opened 2020-10-01
Add retry conditions to volume creation and resizing steps

furkatgofurov7:
When test executor volume creation or resizing it fails, it will stay in an infinitive loop trying to create or resize volume. This PR avoids infinitive looping and adds retrying in case of above mentioned failures.

### DIFF
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -72,34 +72,25 @@ then
 
   # Create test executer volume from copy of base volume
   echo "Creating a test executer volume from copy of base volume."
-  TEST_EXECUTER_VOLUME_ID="$(openstack volume create -f json \
-    --source "${BASE_VOLUME_NAME_ID}" \
-    "${TEST_EXECUTER_VM_NAME}" | jq -r '.id')"
+  create_test_executer_volume "${BASE_VOLUME_NAME_ID}" "${TEST_EXECUTER_VM_NAME}"
   
-  #--size 200 \
-  # Wait for a test executer volume to be available...
+  # Wait for a test executer volume to be available
   echo "Waiting for a test executer volume to be available."
-  until openstack volume show "${TEST_EXECUTER_VM_NAME}" -f json \
-    | jq .status | grep "available"
-  do
-    sleep 10
-  done
-  
+  wait_for_volume "${BASE_VOLUME_NAME_ID}" "${TEST_EXECUTER_VM_NAME}"
+
+  # Resize test executer volume
   echo "Resizing test executer volume."
   openstack volume set --size "${RESIZED_VM_SIZE}" "${TEST_EXECUTER_VM_NAME}"
   
+  # Wait for a resized test executer volume to be available
   echo "Waiting for a resized test executer volume to be available."
-  until openstack volume show "${TEST_EXECUTER_VM_NAME}" -f json \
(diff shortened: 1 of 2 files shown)


kashifest:
looks good at a first glance, I would wait for someone with better knowledge on this to review it.

Jaakko-Os (commented):
lgmt

fmuyassarov:
Sorry if I'm asking basic questions, but how often does the issue happen? Because, I'm wondering if the reducing number of retries would really solve the issue. If it happens for all the current jobs, then they will anyways fail but a bit faster because the number of retries to create/resize the volume is 1 right? Maybe we should look more deeper to understand what is the reason that the volume creation/resizing fails?

furkatgofurov7:
> Sorry if I'm asking basic questions, but how often does the issue happen? Because, I'm wondering if the reducing number of retries would really solve the issue. If it happens for all the current jobs, then they will anyways fail but a bit faster because the number of retries to create/resize the volume is 1 right? Maybe we should look more deeper to understand what is the reason that the volume creation/resizing fails?

It helps to get out of the loop rather than staying indefinitely and fail.

jenkins/scripts/integration_test.sh
```diff
@@ -68,27 +72,60 @@ then
...
+      | jq .status)" == "error"* ]];
+    then
+      echo "Volume resizing is failed"
+      # If volume resizing is failed, then retry volume creation and resizing only once
+      if [ $retry -eq 0 ]; then
+        echo "Deleting a volume that failed to be resized..."
+        openstack volume delete "${TEST_EXECUTER_VM_NAME}"
+        echo "Creating a new test executer volume and resizing..."
+        create_test_executer_volume "${BASE_VOLUME_NAME_ID}" "${TEST_EXECUTER_VM_NAME}"
+        openstack volume set --size "${RESIZED_VM_SIZE}" "${TEST_EXECUTER_VM_NAME}"
```
maelk, line 122: Shouldn't this wait for the volume to be ready first ?

furkatgofurov7: Yeah true, it should wait again before resizing it

furkatgofurov7: Done

jenkins/scripts/integration_test.sh
```diff
@@ -72,27 +72,65 @@ then
...
+      if [ $retry -eq 0 ]; then
+        echo "Deleting a volume that failed to be resized..."
+        openstack volume delete "${TEST_EXECUTER_VM_NAME}"
+        echo "Creating a new test executer volume and resizing..."
+        create_test_executer_volume "${BASE_VOLUME_NAME_ID}" "${TEST_EXECUTER_VM_NAME}"
+        until openstack volume show "${TEST_EXECUTER_VM_NAME}" -f json \
+          | jq .status | grep "available"
+        do
+          sleep 10
+        done
```
maelk, line 126: you probably need an escape condition here if the volume fails to create and ends in error state, to not only rely on the timeout.

furkatgofurov7: Done. I took completely each action of the test_integration script out for better readability. Please have a look @maelk !

maelk (approved):

fmuyassarov (commented):
/lgtm

PR #159, opened 2020-10-05
Fix upgrade test log collection

Jaakko-Os:
The test case result file was missing. This fix
includes it to the tar package.

### DIFF
--- a/jenkins/scripts/files/run_fetch_logs.sh
+++ b/jenkins/scripts/files/run_fetch_logs.sh
@@ -64,7 +64,7 @@ sudo chown -R ${USER}:${USER} "${LOGS_DIR}/qemu"
 if [[ "${TESTS_FOR}" == "feature_tests_upgrade"* ]]
 then
   mkdir -p "${LOGS_DIR}/upgrade"
-  sudo sh -c "cp /tmp/\.*upgrade.result.txt ${LOGS_DIR}/upgrade/"
+  sudo sh -c "cp /tmp/*upgrade.result.txt ${LOGS_DIR}/upgrade/"
   sudo chown -R ${USER}:${USER} "${LOGS_DIR}/upgrade"
 fi
 


fmuyassarov (commented):
In the case of the Ubuntu upgrade test, there is no /upgrade folder. Also, what is the difference between two *upgrade.result.txt files? Are they representing different upgrade tests? If yes, wouldn't it better to rename to unique name in order to not confuse ourselves?

Jaakko-Os:
> In the case of the Ubuntu upgrade test, there is no /upgrade folder in the Jenkins logs. Also, what is the difference between two *upgrade.result.txt files? Are they representing different upgrade tests? If yes, wouldn't it better to rename them to unique names in order to not confuse ourselves?

The issue behind the missing upgrade test result file is the naming:
- the test **output** log file name starts with a  '.'
- the test **result** log file name starts with a number representing the current year,
    * the result file contains a line per run test case: <TC-name> pass|fail

This fix copies the test result file too into the log archive. All that in a single PR.

Why not change the test output file naming (remove the '.')?
That would mean two PRs and verification for both which is very time consuming and difficult in the current CI.

fmuyassarov:
> > In the case of the Ubuntu upgrade test, there is no /upgrade folder in the Jenkins logs. Also, what is the difference between two *upgrade.result.txt files? Are they representing different upgrade tests? If yes, wouldn't it better to rename them to unique names in order to not confuse ourselves?
> 
> The issue behind the missing upgrade test result file is the naming:
> 
>     * the test **output** log file name starts with a  '.'
> 
>     * the test **result** log file name starts with a number representing the current year,
>       
>       * the result file contains a line per run test case: TC-name pass|fail
> 
> 
> This fix copies the test result file too into the log archive. All that in a single PR.
> 
> Why not change the test output file naming (remove the '.')?
> That would mean two PRs and verification for both which is very time consuming and difficult in the current CI.

I just realized that Ubuntu test here is an integration test not the upgrade test. That's why there is no upgrade logs available. Shall we trigger the Ubuntu based upgrade test.

maelk (approved):

Jaakko-Os:
wait until https://github.com/metal3-io/metal3-dev-env/pull/503 is merged

PR #160, opened 2020-10-05
Fix condition checking when volume status contains an error substring

furkatgofurov7:
When resizing a volume, volume still goes into `error_extending` status and stays there due to the improper condition check. This PR fixes that, by checking if volume status contains the substring `error` and also adds 10s delay while deleting a volume for proper deletion.


maelk (approved):

fmuyassarov (commented):
/lgtm

PR #161, opened 2020-10-07
Fix volume resizing in case of feature tests

furkatgofurov7:
This PR fixes volume resizing when we are running feature tests. Currently volume is resized to 100GB for all tests, however feature tests need at least 300GB.

### DIFF
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -38,14 +38,25 @@ NUM_NODES="${NUM_NODES:-2}"
 TESTS_FOR="${TESTS_FOR:-integration_test}"
 
 DEFAULT_HOSTS_MEMORY="${DEFAULT_HOSTS_MEMORY:-4096}"
-RESIZED_VM_SIZE="${RESIZED_VM_SIZE:-100}"
+if [[ "${TESTS_FOR}" == "feature_tests"* ]]
+then
+  RESIZED_VM_SIZE="${RESIZED_VM_SIZE:-300}"
+else
+  RESIZED_VM_SIZE="${RESIZED_VM_SIZE:-100}"
+fi
+
 VM_TIMELABEL="${VM_TIMELABEL:-$(date '+%Y%m%d%H%M%S')}"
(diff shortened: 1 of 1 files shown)


jenkins/scripts/integration_test.sh
```diff
@@ -38,14 +38,25 @@ NUM_NODES="${NUM_NODES:-2}"
...
 TEST_EXECUTER_PORT_NAME="${TEST_EXECUTER_PORT_NAME:-${TEST_EXECUTER_VM_NAME}-int-port}"
 
 if [[ "${TESTS_FOR}" == "feature_tests"* ]]
 then
-  TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-300GB}"
+  if [ "${DISTRIBUTION}" == "ubuntu" ]
+  then
+    TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB}"
+  else
+    TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-300GB}"
```
kashifest, line 58: Why do we need the different flavor for centos and ubuntu?

furkatgofurov7: For Ubuntu, we resize volume to 300GB and create VM from it so flavor is 0GB for local disk, In case of CentOS, we do not have volume, so we just create VM with flavor which has 300GB

furkatgofurov7: Maybe was not clear. In **normal** Ubuntu tests, we resize volume to 100GB and we use flavor 4C-16GB so that 100GB volume will be attached to VM and makes total disk size 100GB, while in CentOS since we do not use volume we just use flavor 4C-16GB-100GB. 

For Ubuntu **feature tests**, we resize volume to 300GB and we use flavor 8C-32GB so that 300GB volume will be attached to VM making total disk size 300 GB, while in CentOS since we do not use volume we just use flavor 8C-32GB-300GB.

kashifest: Aha cool, by the way its not 0 GB, it takes 20GB for local disk

maelk: it's anyways ignored by Openstack at the end AFAIK , so it does not really matter.

PR #164, opened 2020-11-05
Add target cluster logs to all CI run.

smoshiur1237:
This PR will include the target cluster logs for all CI run.

### DIFF
--- a/jenkins/scripts/files/run_fetch_logs.sh
+++ b/jenkins/scripts/files/run_fetch_logs.sh
@@ -68,14 +68,11 @@ then
   sudo chown -R ${USER}:${USER} "${LOGS_DIR}/upgrade"
 fi
 
-if [[ "${TESTS_FOR}" == "feature_tests" || "${TESTS_FOR}" == "feature_tests_centos" ]]
+target_config=$(sudo find /tmp/ -type f -name "kubeconfig*")
+if [ "${target_config}" != "" ]
 then
-  target_config=$(sudo find /tmp/ -type f -name "kubeconfig*")
-  if [ "${target_config}" != "" ]
-  then
-    #fetch target cluster k8s logs
-    fetch_k8s_logs "target_cluster" $target_config
-  fi
+  #fetch target cluster k8s logs
+  fetch_k8s_logs "target_cluster" $target_config
 fi
 
 tar -cvzf "$LOGS_TARBALL" ${LOGS_DIR}/*


fmuyassarov (commented):
For the upgrade we already have a condition to fetch the logs. 

```
if [[ "${TESTS_FOR}" == "feature_tests_upgrade"* ]]
then
  mkdir -p "${LOGS_DIR}/upgrade"
  sudo sh -c "cp /tmp/\.*upgrade.result.txt ${LOGS_DIR}/upgrade/"
  sudo chown -R ${USER}:${USER} "${LOGS_DIR}/upgrade"
fi
```

kashifest (approved):

fmuyassarov:
I'm not sure why, but the k8s_target_cluster folder is empty on the feature test run.

kashifest:
> I'm not sure why, but the k8s_target_cluster folder is empty on the feature test run.

Didnt I explain why it will be empty on a successful run  ?

fmuyassarov:
> > I'm not sure why, but the k8s_target_cluster folder is empty on the feature test run.
> 
> Didnt I explain why it will be empty on a successful run ?

I'm confused now. Because, I thought it would fetch on all the runs. So, we will have the target cluster logs only when CI failed?

kashifest:
> > > I'm not sure why, but the k8s_target_cluster folder is empty on the feature test run.
> > 
> > 
> > Didnt I explain why it will be empty on a successful run ?
> 
> I'm confused now. Because, I thought it would fetch on all the runs. So, we will have the target cluster logs only when CI failed?

By the time fetch log is run, target control plane is already deprovisioned on a successful run.

fmuyassarov:
> > > > I'm not sure why, but the k8s_target_cluster folder is empty on the feature test run.
> > > 
> > > 
> > > Didnt I explain why it will be empty on a successful run ?
> > 
> > 
> > I'm confused now. Because, I thought it would fetch on all the runs. So, we will have the target cluster logs only when CI failed?
> 
> By the time fetch log is run, target control plane is already deprovisioned on a successful run.

Ow, now I remember. So, that should be fixed in the metal3-dev-env right?

kashifest:
> > > > > I'm not sure why, but the k8s_target_cluster folder is empty on the feature test run.
> > > > 
> > > > 
> > > > Didnt I explain why it will be empty on a successful run ?
> > > 
> > > 
> > > I'm confused now. Because, I thought it would fetch on all the runs. So, we will have the target cluster logs only when CI failed?
> > 
> > 
> > By the time fetch log is run, target control plane is already deprovisioned on a successful run.
> 
> Ow, now I remember. So, that should be fixed in the metal3-dev-env right?

Yes. I ll try to add an ansible task at the end of move.yml to copy everything back to source cluster before pivot back.

PR #168, opened 2020-12-04
CI: Run upgrade test with CAPI v0.3.10 release.

smoshiur1237:


namnx228:
lgmt

jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -132,6 +132,7 @@ if [[ "${TESTS_FOR}" == "feature_tests_upgrade"* ]]
 then
   export IRONIC_BASIC_AUTH="false"
   export IRONIC_TLS_SETUP="false"
+  export CAPIRELEASE="v0.3.10"
```
fmuyassarov, line 135: Is it so that we are going to use v`0.3.10` only for the upgrade tests until there is a new release with KCP fix? Also, you will need I think to parametrize  CAPIRELEASE var in Metal3-dev-env so that it accepts the value set in project-infra.

smoshiur1237: Yes when the fix is available, will remove this.

fmuyassarov (approved):
Would you mind please to add a note in the PR description why we need it?

PR #175, opened 2021-03-24
Test sushy-tools/vbmc in metal3-dev-env CI

hardys:
These container images are defined in the metal3-dev-env repo
but currently aren't tested in CI, leading to potential regressions
such as https://github.com/metal3-io/metal3-dev-env/pull/609

We should always build/test these for metal3-dev-env CI, to catch
issues before changes to the Dockerfiles land.

### DIFF
--- a/jenkins/scripts/files/run_integration_tests.sh
+++ b/jenkins/scripts/files/run_integration_tests.sh
@@ -53,6 +53,9 @@ if [ "${REPO_NAME}" == "metal3-dev-env" ]
 then
   export METAL3REPO="${UPDATED_REPO}"
   export METAL3BRANCH="${UPDATED_BRANCH}"
+  # Build/test the sushy-tools/vbmc images, since they are defined in this repo
+  export SUSHY_TOOLS_LOCAL_IMAGE="/home/${USER}/tested_repo/resources/sushy-tools"
+  export VBMC_LOCAL_IMAGE="/home/${USER}/tested_repo/resources/vbmc"
 
   # If the target repo and branch are the same as the source repo and branch
   # we're running a master test, that is not for a PR, so we build the image


fmuyassarov (approved):
Thanks @hardys ! I believe this is enough but there is no way we can test right now if the vmbc and sushy-tools images are going to built correctly and pushed to the local registry in metal3-devp-env. With the current change, we will be able to test only when PR is opened in Metal3-dev-env repo due to` if [ "${REPO_NAME}" == "metal3-dev-env" ]`. However, if we add the same right now for 
`elif [[ "${REPO_NAME}" == "project-infra" ]]` then we could trigger integration CI here and see the results. Would you mind to add the same for under 
`elif [[ "${REPO_NAME}" == "project-infra" ]]` for testing and if okay, then remove it before we merge it?

fmuyassarov:
Sorry for bothering you too much with this, unfortunately there is no any better way to test that I can see.

hardys:
> Sorry for bothering you too much with this, unfortunately there is no any better way to test that I can see.

@fmuyassarov np!  I just pushed a temporary commit adding to the project-infra repo, is `test-integration` enough to trigger the test?  I didn't see it run anything yesterday...

fmuyassarov:
> > Sorry for bothering you too much with this, unfortunately there is no any better way to test that I can see.
> 
> @fmuyassarov np! I just pushed a temporary commit adding to the project-infra repo, is `test-integration` enough to trigger the test? I didn't see it run anything yesterday...

Thank you. Not sure what happened to the CI, I just triggered it, and let's see how it goes.

jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -104,6 +107,10 @@ then
 
 elif [[ "${REPO_NAME}" == "project-infra" ]]
 then
+  # FIXME(hardys) only for testing
+  # Build/test the sushy-tools/vbmc images, since they are defined in this repo
+  export SUSHY_TOOLS_LOCAL_IMAGE="/home/${USER}/tested_repo/resources/sushy-tools"
+  export VBMC_LOCAL_IMAGE="/home/${USER}/tested_repo/resources/vbmc"
```
fmuyassarov, line 113: @hardys Can you please try with these changes, because it seems we are not giving the right path here. In this scenario, Metal3-dev-env is cloned to [`metal3`](https://github.com/metal3-io/project-infra/blob/c3e4635fef344c839516b068abd52a3dc0d78e2b/jenkins/scripts/files/run_integration_tests.sh#L126)
```suggestion
  export SUSHY_TOOLS_LOCAL_IMAGE="metal3/resources/sushy-tools"
  export VBMC_LOCAL_IMAGE="metal3/resources/vbmc"
```

fmuyassarov:
I have tested these changes https://github.com/metal3-io/metal3-dev-env/pull/618, and the are ok to merge, as I can see both vbmc and sushy-tools container images were built as expected in https://jenkins.nordix.org/job/airship_metal3io_metal3_dev_env_v1a4_integration_test_ubuntu/451/.

/lgtm 
/hold cancel

PR #178, opened 2021-04-12
Remove IPAM jobs from prow temporarily

kashifest:
We are making the ipam jobs optional temporarily since prow is WIP and we are relying on travis for the time being. 

Reference: https://github.com/metal3-io/ip-address-manager/pull/27

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -559,6 +559,7 @@ presubmits:
   - name: gofmt
     always_run: true
     decorate: true
+    optional: true
     spec:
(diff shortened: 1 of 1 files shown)


prow/config/config.yaml
```diff
@@ -556,104 +556,104 @@ presubmits:
         imagePullPolicy: Always
 
   metal3-io/ip-address-manager:
```
fmuyassarov, line 558: I'm not sure, but maybe we should comment out this line too.

PR #181, opened 2021-04-21
Get Ironic Python Agent from Artifactory

macaptain:
In order to reduce dependency on upstream services (e.g. in the event of network degradation between Sweden and US West, where the default Ironic Python Agent tarball is hosted), set IPA_BAREURI to download IPA from Nordix Artifactory in CI. 

This has the added benefit of potentially reducing build times, because we'll be downloading from the same infrastructure on which the CI is hosted.


maelk (approved):

fmuyassarov (approved):

smoshiur1237 (approved):
/lgtm

PR #184, opened 2021-05-04
Fetch also exited container logs

kashifest:
This PR 
1. makes sure that when we fetch the container logs we also take the dead container logs since those can also give information about why a container might be failing. Currently we always have to login to the
CI vm to find out whats happening in-case an ironic container did not start properly.
2. Removes unnecessary image building for CAPM3 and BMO on a Project-infra integration test.


fmuyassarov (commented):
/lgtm

PR #185, opened 2021-05-06
Remove duplication in external plugins

fmuyassarov:
We have duplicated [lines of code](https://github.com/metal3-io/project-infra/blob/master/prow/config/plugins.yaml#L47-L65) for enabling external plugin needs-rebase. This patch removes the duplication for that.


maelk:
Please add details. What is it duplicating ?

fmuyassarov:
> Please add details. What is it duplicating ?

Done.

maelk:
/lgtm
to get the fix in asap

PR #186, opened 2021-05-06
Don't use prow image autobump

fmuyassarov:
I think even in the previous prow setup we weren't using autobump, and by looking at [past PRs](https://github.com/metal3-io/project-infra/pulls?q=is%3Apr+author%3Ametal3-io-bot+is%3Aclosed), it seems automatic image updates didn't go well. I've also looked at other repos like KubeVirt, Jetstack, Falco and it seems they are also doing image updates manually.
I can take care of updating those images from time to time. The good thing from doing updates manually is sometimes we might need to make some configuration changes in our manifests/config/plugins files before taking the latest images.

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -191,57 +191,6 @@ periodics:
      - name: token
        secret:
          secretName: github-token
-#  - name: ci-prow-autobump
-#    # https://github.com/kubernetes/test-infra/tree/master/prow/cmd/autobump#prow-autobump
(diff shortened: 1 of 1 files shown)


kashifest:
/lgtm
What about adding renovate here ?

russellb:
/approve
/lgtm

I liked the idea of this.  It "worked" in that it did create PRs for me, but by the time I set this up, prow had changed too much from what I was running to make the upgrade that simple.

PR #190, opened 2021-05-13
disable automatic self-approval and ensure 2 lgtm does not approve

dhellmann:
As a community-driven open source project, we want all PR authors to
be subject to the same review requirements. That includes requiring
folks who have approval rights to have 2 reviews, including a review
from another approver. To achieve this, enable the `approve` plugin
setting `require_self_approval` to require explicit approval of PRs.

At the same time, ensure that all PRs are examined by someone with
approval permission by explicitly disabling `lgtm_acts_as_approve`.

Discussion on the mailing list: https://groups.google.com/g/metal3-dev/c/H9rR-mjgJ5s

### DIFF
--- a/prow/config/plugins.yaml
+++ b/prow/config/plugins.yaml
@@ -44,6 +44,17 @@ plugins:
     plugins:
     - config-updater
 
+approve:
+  - repos:
+    - metal3-io
+
+    # RequireSelfApproval requires PR authors to explicitly approve their PRs.
+    # Otherwise the plugin assumes the author of the PR approves the changes in the PR.
+    require_self_approval: true
+
+    # A /lgtm from a single approver should not allow a PR to merge.
+    lgtm_acts_as_approve: false
+
 external_plugins:
   metal3-io:
   - name: needs-rebase


dhellmann:
I based this PR on the instructions in https://prow.k8s.io/plugins and a similar setting we have for OpenShift (https://github.com/openshift/release/blob/master/core-services/prow/02_config/_plugins.yaml#L73), but apparently I got it wrong. Maybe someone more familiar with Prow can help me figure out what I've done wrong?

dhellmann:
Let's see if just making it a list works...

dhellmann:
> Let's see if just making it a list works...

Seems like it.

zaneb:
Mailing list thread: https://groups.google.com/g/metal3-dev/c/H9rR-mjgJ5s
/approve

prow/config/plugins.yaml
```diff
@@ -44,6 +44,17 @@ plugins:
...
+approve:
+  - repos:
+    - metal3-io
+
+    # RequireSelfApproval requires PR authors to explicitly approve their PRs.
+    # Otherwise the plugin assumes the author of the PR approves the changes in the PR.
+    require_self_approval: true
+
+    # A /lgtm from a single approver should not allow a PR to merge.
+    lgtm_acts_as_approve: false
```
fmuyassarov, line 56: Wasn't this already the case by default ? but +1 to explicitly disable it.

fmuyassarov (commented):
/approve

prow/config/plugins.yaml
zaneb: It's possible the default changed when you upgraded the prow cluster, but as of a few weeks ago it was definitely enabled. See e.g. https://github.com/metal3-io/baremetal-operator/pull/852 where the bot adds the approved label as soon as the PR is opened.

maelk:
I would also recommend to switch the reviews process, start by putting lgtm on the PRs, and approve only once a PR has a lgtm, that puts the final word on the approver. That is also because lgtm is removed when the pr is rebased, or the commits change in any way, while the approval remains.

prow/config/plugins.yaml
fmuyassarov: I see, then I misunderstood `lgtm_acts_as_approve` a bit, because I though adding lgtm would even approve the PR. It is true that if you are an approver in `OWNERS` file, then you will get approved label immediately when you open a PR and it was the same from the beginning of the usage of Prow.

dhellmann: There are 2 aspects of this and they're closely related. I may not have clearly explained things in the comments, so let me try to clarify here and then we can update the file with a better description.

The default for `require_self_approval` is false, which means then someone in the approver list submits a PR that PR is automatically approved. The effect of that is that they then only need 1 person to leave a `/lgtm` comment for the PR to merge. We want it set to true so all PRs require an explicit `/approve` comment as well.

The default for `lgtm_acts_as_approve` is false, which means that an approver who comments `/lgtm` does not also implicitly add the approved label. We do not want the value set to true, because that would also introduce a path for a PR to be merged by a single person. It's redundant, but I'm setting it explicitly to false here so the change in the file will be associated with the git commit message linking to the discussion of why we want it that way.

fmuyassarov: Got it, thanks for explanation @dhellmann

zaneb: Sorry, I got confused about which line this comment was referring to :)

I thought that we had `lgtm_acts_as_approve` on as well (which is why I mostly avoid doing /lgtm), but perhaps that was on our downstream repo.

fmuyassarov:
This was brought in the community meeting today and there were no objection on this anymore. Merging it.
/lgtm

PR #198, opened 2021-06-03
Add .cluster-api config to fetch logs tarball

macaptain:
Configuration is written to ~/.cluster-api, edited and reapplied during the upgrade tests. It would be useful to examine the contents when debugging.

### DIFF
--- a/jenkins/scripts/files/run_fetch_logs.sh
+++ b/jenkins/scripts/files/run_fetch_logs.sh
@@ -61,6 +61,9 @@ mkdir -p "${LOGS_DIR}/qemu"
 sudo sh -c "cp -r /var/log/libvirt/qemu/* ${LOGS_DIR}/qemu/"
 sudo chown -R ${USER}:${USER} "${LOGS_DIR}/qemu"
 
+mkdir -p "${LOGS_DIR}/cluster-api-config"
+cp -r "/home/airshipci/.cluster-api/." "${LOGS_DIR}/cluster-api-config/"
+
 if [[ "${TESTS_FOR}" == "feature_tests_upgrade"* ]]
 then
   mkdir -p "${LOGS_DIR}/upgrade"


lentzi90 (approved):
Approving, and I guess that is all I can do. :slightly_smiling_face:

kashifest (approved):

maelk:
/test-integration
Please check if this is in the tarball as expected

jenkins/scripts/files/run_fetch_logs.sh
```diff
@@ -61,6 +61,9 @@ mkdir -p "${LOGS_DIR}/qemu"
 sudo sh -c "cp -r /var/log/libvirt/qemu/* ${LOGS_DIR}/qemu/"
 sudo chown -R ${USER}:${USER} "${LOGS_DIR}/qemu"
 
+mkdir -p "${LOGS_DIR}/cluster-api-config"
+cp -r "/home/airshipci/.cluster-api/*" "${LOGS_DIR}/cluster-api-config/"
```
wgslr, line 65:
```suggestion
cp -r "/home/airshipci/.cluster-api/." "${LOGS_DIR}/cluster-api-config/"
```
This will also catch hidden files. Maybe not particularly likely in ~/.cluster-api, but seems to me a better general practice to get an accurate dump.

macaptain:
> /test-integration
> Please check if this is in the tarball as expected

The cluster-api-config directory is in the tarball. It's empty, but I don't expect integration tests to run `clusterctl` commands, so I need to run an upgrade test in order to see a change. Can we merge this, I'll trigger the upgrade test and confirm that I have the config I need to debug the failures that we are currently seeing in the upgrade test?

macaptain:
Confirmed that I have the logs that I want in the test-upgrade-features logs tarball archive.

wgslr (approved):

kashifest:
/lgtm
taking this in since it only adds value to the logs.

PR #208, opened 2021-06-16
Handle e2e tests in integration test scripts

wgslr:
Run the CAPM3 e2e tests on Jenkins.

Relies on [CAPM3#229](https://github.com/metal3-io/cluster-api-provider-metal3/pull/229)

### DIFF
--- a/jenkins/scripts/files/run_integration_tests.sh
+++ b/jenkins/scripts/files/run_integration_tests.sh
@@ -140,6 +140,11 @@ then
 elif [[ "${TESTS_FOR}" == "feature_tests" || "${TESTS_FOR}" == "feature_tests_centos" ]]
 then
   make feature_tests
+elif [[ "${TESTS_FOR}" == "e2e_tests" ]]
+then
+  pushd "${CAPM3PATH}"
+  make test-e2e
+  popd
 else
   make
   make test
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -49,7 +49,7 @@ VM_TIMELABEL="${VM_TIMELABEL:-$(date '+%Y%m%d%H%M%S')}"
 TEST_EXECUTER_VM_NAME="${TEST_EXECUTER_VM_NAME:-ci-test-vm-${VM_TIMELABEL}}"
 TEST_EXECUTER_PORT_NAME="${TEST_EXECUTER_PORT_NAME:-${TEST_EXECUTER_VM_NAME}-int-port}"
 
-if [[ "${TESTS_FOR}" == "feature_tests"* ]]
+if [[ "${TESTS_FOR}" == "feature_tests"* || ${"TEST_FOR"} == "e2e_tests"* ]]
 then
   if [ "${DISTRIBUTION}" == "ubuntu" ]
   then


jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -140,6 +140,9 @@ then
 elif [[ "${TESTS_FOR}" == "feature_tests" || "${TESTS_FOR}" == "feature_tests_centos" ]]
 then
   make feature_tests
+elif [[ "${TESTS_FOR}" == "e2e_tests" ]]
+then
+  make test-e2e
```
fmuyassarov, line 145: I think you need to cd/pushd to local CAPM3 path in order to call `test-e2e` Makefile target of CAPM3 repo. Because, at this point we are inside l[ocal copy of metal3-dev-env](https://github.com/metal3-io/project-infra/blob/master/jenkins/scripts/files/run_integration_tests.sh#L127-L134) repository.

wgslr: Good point, fixed

jenkins/scripts/integration_test.sh
```diff
@@ -49,7 +49,7 @@ VM_TIMELABEL="${VM_TIMELABEL:-$(date '+%Y%m%d%H%M%S')}"
 TEST_EXECUTER_VM_NAME="${TEST_EXECUTER_VM_NAME:-ci-test-vm-${VM_TIMELABEL}}"
 TEST_EXECUTER_PORT_NAME="${TEST_EXECUTER_PORT_NAME:-${TEST_EXECUTER_VM_NAME}-int-port}"
 
-if [[ "${TESTS_FOR}" == "feature_tests"* ]]
+if [[ "${TESTS_FOR}" == "feature_tests"* || ${"TEST_FOR"} == "e2e_tests*" ]]
```
furkatgofurov7, line 52: should not this be:
```suggestion
if [[ "${TESTS_FOR}" == "feature_tests"* || ${"TEST_FOR"} == "e2e_tests"* ]]
```
?

The meaning of "*" wildcard in case of `feature_tests` was if it matches substring(feature_tests) of string(i.e feature_tests_centos), it will meet the requirement, meaning if we will have `e2e_tests` for more than one flavor we could do the same here.

furkatgofurov7 (commented):
Depending on the how long will be enough to run the e2e tests, you might need to update [pipeline](https://github.com/metal3-io/project-infra/blob/master/jenkins/jobs/integration_tests.pipeline#L9)

wgslr:
> Depending on the how long will be enough to run the e2e tests, you might need to update [pipeline](https://github.com/metal3-io/project-infra/blob/master/jenkins/jobs/integration_tests.pipeline#L9)

On my VM the time needed is ~1h. So the default timeout of 5400 _should_ be enough... but I can make it 18000 to be on the safe side.

jenkins/scripts/integration_test.sh
wgslr: Yes, of course :sweat_smile:

wgslr: Updated

furkatgofurov7:
> On my VM the time needed is ~1h. So the default timeout of 5400 _should_ be enough... but I can make it 18000 to be on the safe side.

We made feature tests 5h because default 1,5h was not enough. If latter is enough for running e2e no need to make it run the same amount of time as feature test.

jenkins/jobs/integration_tests.pipeline
```diff
@@ -6,7 +6,7 @@ def CLEAN_TIMEOUT = 600
 def TIMEOUT = 5400
 
 if (env.TESTS_FOR) {
-  if ( (env.TESTS_FOR).startsWith('feature_tests') ) {
+  if ( (env.TESTS_FOR).startsWith('feature_tests') || (env.TESTS_FOR).startsWith('e2e_tests')  ) {
```
furkatgofurov7, line 9: no need for this change

jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -140,6 +140,11 @@ then
 elif [[ "${TESTS_FOR}" == "feature_tests" || "${TESTS_FOR}" == "feature_tests_centos" ]]
 then
   make feature_tests
+elif [[ "${TESTS_FOR}" == "e2e_tests" ]]
+then
+  pushd "/home/${USER}/tested_repo"
```
fmuyassarov, line 145: you have a var, why not to use it 😊
```suggestion
  pushd "${CAPM3PATH}"
```

wgslr: Done

fmuyassarov (commented):
/approve

furkatgofurov7 (commented):
/lgtm

PR #212, opened 2021-06-22
Update DockerHub registry path

macaptain:
The registry path for DockerHub https://registry.hub.docker.com/v2/ returns a 404. Docker/Podman make a GET request to this path before pulling the image, and they're failing at this step. Whether DockerHub have deprecated or changed the behaviour of their registry is unclear.

Updating the paths to docker.io resolves this issue and allows Docker image pulling once more. This should fix the Prow jobs that are failing markdownlint.

Additionally, images which were being pulled from DockerHub have now been explicitly marked as being pulled from docker.io.

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -207,7 +207,7 @@ presubmits:
         env:
         - name: IS_CONTAINER
           value: "TRUE"
-        image: registry.hub.docker.com/library/golang:1.16
+        image: docker.io/golang:1.16
         imagePullPolicy: Always
   - name: gosec
     always_run: true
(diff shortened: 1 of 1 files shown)


lentzi90 (approved):
LGTM!
For reference I managed to find a reference to `docker.io` in the docs also, and it seems to be the correct URL to use.

> `docker pull ubuntu` instructs docker to pull an image named ubuntu from the official Docker Hub. This is simply a shortcut for the longer `docker pull docker.io/library/ubuntu` command

From https://docs.docker.com/registry/introduction/

PR #225, opened 2021-09-03
append variables to vars.sh file without emptying the file 

Sunnatillo:
While passing variables to metal3-dev-env with file vars.sh, it keeps variables which were passed through Jenkins Pipeline.


Rozzii (approved):

furkatgofurov7 (commented):
/lgtm

PR #226, opened 2021-09-06
Fix the location of k8s version export

kashifest:
The location of k8s version export is not correct. It has to be done in the beginning.


fmuyassarov (approved):

furkatgofurov7 (commented):
/lgtm

PR #236, opened 2021-10-04
Run feature test in Frankfurt region

kashifest:
This PR makes feature test run in Frankfurt region, the reason being checking if it helps to overcome the webhook related issue.


jenkins/scripts/integration_test.sh
```diff
@@ -122,6 +124,24 @@ fi
 TEST_EXECUTER_IP="$(openstack port show -f json "${TEST_EXECUTER_PORT_NAME}" \
   | jq -r '.fixed_ips[0].ip_address')"
 
+if [ "$OS_REGION_NAME" == "Fra1" ]
```
macaptain, line 127: Nit: It's more consistent to use `[[` everywhere instead of `[` if you don't need backwards compatibility with old shells. `[` is fine in this case, but it can be confusing because, for example, pattern matching only works with `[[` not with `[`.

jenkins/scripts/integration_test.sh
```diff
@@ -122,6 +124,24 @@ fi
...
+then 
+  # Fetch Free floating IP 
+  FLOATING_IP="$(openstack floating ip list --status DOWN -c "Floating IP Address" -f value | head -1 )"
+
+  if [ -z $FLOATING_IP ]
```
macaptain, line 132: Either `[[` is needed or `$FLOATING_IP` should be quoted. I don't think we run `shellcheck` on this script (we probably should?) but shellcheck will fail here.

macaptain (commented):
Great stuff! I pointed out two minor bash things.

jenkins/scripts/integration_test.sh
kashifest: That was failing so I had to change it to single `[`

jenkins/scripts/fetch_logs.sh
```diff
@@ -19,10 +19,25 @@ source "${CI_DIR}/utils.sh"
...
+fi
+
 # Get the IP
 TEST_EXECUTER_IP="$(openstack port show -f json "${TEST_EXECUTER_PORT_NAME}" \
   | jq -r '.fixed_ips[0].ip_address')"
 
+if [[ "${TESTS_FOR}" == "feature_tests"* ]]
+then
+  OS_REGION_NAME="Fra1"
+  OS_AUTH_URL="https://fra1.citycloud.com:5000"
```
furkatgofurov7, line 35: why we need to check the same condition twice? can't we have it all in one condition?

jenkins/scripts/integration_test.sh
```diff
@@ -50,14 +50,16 @@ VM_TIMELABEL="${VM_TIMELABEL:-$(date '+%Y%m%d%H%M%S')}"
...
 if [[ "${TESTS_FOR}" == "feature_tests"* || "${TESTS_FOR}" == "e2e_tests"* ]]
 then
-  if [ "${DISTRIBUTION}" == "ubuntu" ]
-  then
-    TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB}"
-  else
-    TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-300GB}"
-  fi
+    # Setting this to 8C-32GB-400GB since other flavors are unavailable in Frankfurt region
+    TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-400GB}"
```
furkatgofurov7, line 62: So, there will not be any volume being used in this region? we had `8C-32GB` for Ubuntu since we attach volume to it and `8C-32GB-300GB` for CentOS where volume is out of picture.

furkatgofurov7 (commented):
Some minor comments

jenkins/scripts/fetch_logs.sh
kashifest: We need the check twice since we do another operation in between. But the env vars need not be set twice. I ll remove it.

jenkins/scripts/integration_test.sh
kashifest: There is no `8C-32GB` flavor in Frankfurt. Ubuntu does create the volume similar to what is done previously. Check the current runs.

jenkins/scripts/integration_test.sh
```diff
@@ -121,6 +123,24 @@ fi
...
+    exit 1
+  fi
+  TEST_EXECUTER_IP="$FLOATING_IP"
+  
+  # Attach floating IP
+  openstack server add floating ip \
+  "${TEST_EXECUTER_VM_NAME}" \
+  "$FLOATING_IP"
+fi
+
```
fmuyassarov, line 143: some explanation on this addition would help to understand this part

namnx228 (commented):
LGTM

jenkins/scripts/integration_test.sh
namnx228: I think this addition solves the issue that a Jenkins worker in Kna1 region cannot connect to VMs in Fra1 because of the different subnets. Therefore, we need to add floating IP to the VM in Fra1, so the worker in Kna1 can talk to them.

fmuyassarov: thank you @kashifest

PR #238, opened 2021-10-06
fix passing  credentials to Bare Metal Lab integration test

Sunnatillo:
fix passing BML_ILO_USERNAME and BML_ILO_PASSWORD credentials to Bare Metal Lab integration test


jenkins/scripts/bml_integration_test.sh
```diff
@@ -69,8 +69,8 @@ ssh \
   -o ServerAliveCountMax=10 \
   -i "${AIRSHIP_CI_USER_KEY}" \
   "${AIRSHIP_CI_USER}"@"${TEST_EXECUTER_IP}" \
-  -o SendEnv="MYVAR" \
-  -o SendEnv="MYVAR" \
+  -o SendEnv="BML_ILO_USERNAME=${BML_ILO_USERNAME}" \
```
macaptain, line 72: I don't think you need the `=` and right hand side. The value of BML_ILO_USERNAME is already set in the environment where this is run.

Sunnatillo:
test-integration-bml-centos

macaptain:
/lgtm

Let's also check that the integration test on VMs still works:

/test-integration

PR #241, opened 2021-10-08
Remove duplicating owners

fmuyassarov:
Remove duplicating owners. Since @kashifest
@furkatgofurov7 and @maelk are now in the root
OWNERS, it is redundant to keep them in the sub
OWNERS too.


fmuyassarov:
> We should most probably remove the file itself? Otherwise this lgtm.

I wanted to keep in case we want to add some reviwers/maintainers in the future. But sure, why not

PR #250, opened 2021-10-18
Run all main builds on Fra1

macaptain:
The Kna1 region isn't as stable as Fra1 and sometimes we see higher CPU steal. We saw improvements moving feature tests to Fra1 region, so let's try moving all builds of the main branch to Fra1 as well. Jobs which are triggered on PRs are still run on Kna1.


macaptain:
> Are we missing fetching logs part as well?:
> 
> https://github.com/metal3-io/project-infra/blob/9542853cf66b6c288d278bb59ed21dd332cda4ee/jenkins/scripts/fetch_logs.sh#L22-L26

Yes, thanks! Fixed.

PR #262, opened 2021-11-01
Don't enable push-gateway in ghProxy

fmuyassarov:
We don't setup prometheus, so there is no benefit of passing these
flags to enable metrics collection.

### DIFF
--- a/prow/manifests/starter.yaml
+++ b/prow/manifests/starter.yaml
@@ -1022,8 +1022,6 @@ spec:
         args:
         - --cache-dir=/cache
         - --cache-sizeGB=99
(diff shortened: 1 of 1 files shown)


kashifest:
This is for prow only right? It wont hamper what @jan-est is doing in m3-dev-env I guess?
/approve

fmuyassarov:
> This is for prow only right? It wont hamper what @jan-est is doing in m3-dev-env I guess? /approve

No it won't. This is internal to Prow, while Jan is working on dev-env related addition.

furkatgofurov7 (commented):
/lgtm

PR #267, opened 2021-11-09
Make v1a5 centos job required in m3-dev-env repo

furkatgofurov7:
Since we migrated to v1b1 now, we should make v1a5 tests as required in m3-dev-env repo replacing v1a4, where latter can be triggered case by case scenarios and we keep only master jobs to watch it.


kashifest (approved):

fmuyassarov (commented):
I hope it won't break the Prow :)
Btw, you need to update the GitHub configuration
too in metal3-dev-env to make it this job required.
/lgtm

PR #269, opened 2021-11-11
Avoid unnecessary CI jobs

zaneb:
Skip CI jobs that are not needed when only e.g. docs or OWNERS files are
modified in the PR.

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -207,7 +207,7 @@ periodics:
 presubmits:
   metal3-io/baremetal-operator:
   - name: gofmt
-    always_run: true
+    run_if_changed: '\.go$'
     decorate: true
     spec:
       containers:
@@ -221,7 +221,7 @@ presubmits:
         image: docker.io/golang:1.16
         imagePullPolicy: Always
   - name: gosec
-    always_run: true
+    skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
     decorate: true
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


prow/config/config.yaml
```diff
@@ -727,7 +716,7 @@ presubmits:
         image: quay.io/metal3-io/capm3-unit:main
         imagePullPolicy: Always
   - name: manifestlint
-    always_run: true
+    skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
```
fmuyassarov, line 719: probably including .go also makes sense here, because it checks only YAML files. Or we could be more explicit with  `run_if_changed:` to run only when YAML/YML files are touched.

prow/config/config.yaml
```diff
@@ -511,7 +505,7 @@ presubmits:
 
   metal3-io/project-infra:
   - name: check-prow-config
-    always_run: true
+    skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
```
fmuyassarov, line 508: `.sh` can also be added here

fmuyassarov (commented):
/approve

prow/config/config.yaml
zaneb: Thanks, I wasn't sure if that would catch everything or not.
One slight worry is that ideally we should run the job if the Makefile or the script it is running changes, and setting that up is ugly given the terrible syntax of these config options.

fmuyassarov: yeah, regex makes it hard to read :(

macaptain:
/lgtm
/hold

@zaneb feel free to unhold whenever you're happy to merge :)

PR #271, opened 2021-11-12
Add release to master CI tests files

smoshiur1237:
This PR will add the release check for FRA1 and add release in related files.


mboukhalfa (approved):

macaptain:
/lgtm

Brilliant, thanks. I think we should extract this logic into a separate function at some point since we've stubbed our toe on this at least twice, forgetting to update this region mapping in some place. Let's follow up on that for the future.

furkatgofurov7 (approved):
/approve

furkatgofurov7:
/retitle Add release branch to cleaning and log fetching scripts

PR #273, opened 2021-11-12
Fix integration cleanup file to add cleanup step for FRA1

smoshiur1237:
This PR will modify the #271 to set it correctly.


furkatgofurov7 (approved):
/approve

PR #274, opened 2021-11-14
Re-configure label synchronization CronJob

fmuyassarov:
This patch moves label synchronization CronJob to default
namespace, thus updates the Makefile entry related to it.
Also, it makes it to reuse the existing github-token Secret,
and bumps the label synchronization image to
[v20211111-bce61c7c4a]( gcr.io/k8s-prow/label_sync:v20211111-bce61c7c4a).

### DIFF
--- a/prow/Makefile
+++ b/prow/Makefile
@@ -9,7 +9,7 @@ update-plugins:
 	kubectl create configmap plugins -n prow --from-file=plugins.yaml=config/plugins.yaml --dry-run=client -o yaml | kubectl replace configmap -n prow plugins -f -
 
 update-labels:
-	kubectl create configmap label-config -n prow --from-file=labels.yaml=labels.yaml --dry-run=client -o yaml | kubectl replace configmap -n prow label-config -f -
+	kubectl create configmap label-config --from-file=labels.yaml=config/labels.yaml --dry-run=client -o yaml | kubectl replace configmap label-config -f -
 
 validate:
 	$(CONTAINER_RUNTIME) run --rm \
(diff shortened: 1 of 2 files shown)


kashifest:
/approve
Do we need 5 commits or squashing them would be a good idea?

smoshiur1237 (approved):
LGTM

fmuyassarov:
> /approve Do we need 5 commits or squashing them would be a good idea?

done

prow/Makefile
```diff
@@ -9,7 +9,7 @@ update-plugins:
 	kubectl create configmap plugins -n prow --from-file=plugins.yaml=config/plugins.yaml --dry-run=client -o yaml | kubectl replace configmap -n prow plugins -f -
 
 update-labels:
-	kubectl create configmap label-config -n prow --from-file=labels.yaml=labels.yaml --dry-run=client -o yaml | kubectl replace configmap -n prow label-config -f -
+	kubectl create configmap label-config --from-file=labels.yaml=labels.yaml --dry-run=client -o yaml | kubectl replace configmap label-config -f -
```
furkatgofurov7, line 12: this should be:
```suggestion
	kubectl create configmap label-config --from-file=labels.yaml=config/labels.yaml --dry-run=client -o yaml | kubectl replace configmap label-config -f -
```
?

furkatgofurov7 (changes requested):

prow/Makefile
fmuyassarov: ops, good catch

fmuyassarov: @furkatgofurov7 thank you, fixed it.

furkatgofurov7 (commented):
/lgtm

PR #279, opened 2021-11-16
Fix unbound prometheus monitoring variable issue

Sunnatillo:
Default value for prometheus_monitoring varialbe set only for feature tests in JJB. There were some tests failed with error unbound prometheus monitoring variable. Here we set value as a false if it was not set above.


jenkins/scripts/integration_test.sh
```diff
@@ -37,6 +37,7 @@ CAPM3_VERSION="${CAPM3_VERSION:-v1beta1}"
 NUM_NODES="${NUM_NODES:-2}"
 TESTS_FOR="${TESTS_FOR:-integration_test}"
 BARE_METAL_LAB=false
+PROMETHEUS_MONITORING="${PROMETHEUS_MONITORING:-false}"
```
fmuyassarov, line 41: don't we have its default value set in JJB?

Sunnatillo: only for feature test. In our case airship dev tools integration test, it is not set. That`s why I was thinking we set false value if it is not set.

fmuyassarov: yep makes sense.

PR #281, opened 2021-11-22
Shrink syslog file size in BML 

Sunnatillo:
This PR adds ansible task which shrinks size of syslog before starting the test in BML .

In Bare metal lab syslog file size sometimes increases up to 20-30 Gb per day. Before running tests on BML we should shrink the size of syslog file in order to avoid failure due to shortage of space

### DIFF
--- a/jenkins/scripts/bare_metal_lab/deploy-lab.yaml
+++ b/jenkins/scripts/bare_metal_lab/deploy-lab.yaml
@@ -41,6 +41,14 @@
       become: true
       when: sshd_config.changed
 
+    - name: Shrink size of syslog
+      become: true
+      become_user: root
+      shell:
+        cmd: truncate --size 10737418240 /var/log/syslog && truncate --size 10737418240 /var/log/syslog.1   2>/dev/null
+      ignore_errors: true
+      tags: cleanup
+
     - name: Clone the metal3-dev-env repo
       git:
         repo: https://github.com/metal3-io/metal3-dev-env.git


jenkins/scripts/bare_metal_lab/deploy-lab.yaml
```diff
@@ -41,6 +41,13 @@
       become: true
       when: sshd_config.changed
 
+    - name: Shrink size of syslog
+      become: true
+      become_user: root
+      shell:
+        cmd: truncate --size 10737418240 syslog.1 && truncate --size 10737418240 syslog  2>/dev/null
```
macaptain, line 48: Using `truncate` is exactly right for this task. However, I don't this will work without the full path to syslog (/var/log/syslog). Otherwise how do we know where the syslog file is? This makes me think that this change has not been manually tested. Could you try running the playbook and let me know if the task runs and if there are any changes needed? (Also - what if syslog.1 doesn't exist? It doesn't have to, maybe we don't have log rotation set up, or maybe the log hasn't gotten big enough to be rotated. Perhaps we should have some tolerance of failure on this command.)

macaptain (changes requested):

jenkins/scripts/bare_metal_lab/deploy-lab.yaml
```diff
@@ -41,6 +41,14 @@
       become: true
       when: sshd_config.changed
 
+    - name: Shrink size of syslog
+      become: true
+      become_user: root
+      shell:
+        cmd: truncate --size 10737418240 /var/log/syslog && truncate --size 10737418240 /var/log/syslog1   2>/dev/null
```
macaptain, line 48: The rotated file is called /var/log/syslog.1, not /var/log/syslog1. How have you tested this command?

macaptain (changes requested):

jenkins/scripts/bare_metal_lab/deploy-lab.yaml
Sunnatillo: I changed the type and tested. It is working now. 
This is the resulted syslog size:    10738MB Dec  8 11:07 syslog

jenkins/scripts/bare_metal_lab/deploy-lab.yaml
Sunnatillo: I tested a playbook after addressing your comments. It is working.  Is the following filed enough for failure tolerance? 
ignore_errors: true

Sunnatillo:
> @Sunnatillo Is it normal that it is generating 20-30 Gb of logs per day? That seems quite big

when test fails at some point in BM LAB dnsmasq creates a lot of logs.

macaptain:
> > @Sunnatillo Is it normal that it is generating 20-30 Gb of logs per day? That seems quite big
> 
> when test fails at some point in BM LAB dnsmasq creates a lot of logs.

I think we should aim to tackle this as well and turn down dnsmasq's logging level. However, this looks like a workaround which will prevent the job from failing.

Thanks for testing the new task, it's always important to try out the task / command.

/lgtm

fmuyassarov:
`--quiet-dhcp, --quiet-dhcp6, --quiet-ra` flags might help to reduce some of the logs. The good thing with this is, we still will have error logs if any.
Reference: https://manpages.ubuntu.com/manpages/jammy/en/man8/dnsmasq.8.html
/approve

PR #282, opened 2021-11-25
Fix syntax issues in log fetching script

Rozzii:
Now the script passes shellcheck.

### DIFF
--- a/jenkins/scripts/files/run_fetch_logs.sh
+++ b/jenkins/scripts/files/run_fetch_logs.sh
@@ -3,7 +3,7 @@
 #Do not fail on error (for example k8s cluster not available)
 set -u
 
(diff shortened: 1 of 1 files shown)


macaptain:
Nice! Does this script pass shellcheck now @Rozzii ? If so, could you update the description to say that. Let's run integration test once to check we fetch the logs correctly:

/test-integration

Rozzii:
> Nice! Does this script pass shellcheck now @Rozzii ? If so, could you update the description to say that. Let's run integration test once to check we fetch the logs correctly:
> 
> /test-integration

Yes it does, I have updated the description, but it looks like /test-integration command is not working for this repo.

furkatgofurov7 (approved):
Last run output looks good. 

/approve

PR #283, opened 2021-11-30
Increase VM storage volume size

Rozzii:
This commit was created because out of space erros have been
noticed during CI job execution.

### DIFF
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -40,12 +40,6 @@ BARE_METAL_LAB=false
 PROMETHEUS_MONITORING="${PROMETHEUS_MONITORING:-false}"
 
 DEFAULT_HOSTS_MEMORY="${DEFAULT_HOSTS_MEMORY:-4096}"
-if [[ "${TESTS_FOR}" == "feature_tests"* || "${TESTS_FOR}" == "e2e_tests"* ]]
-then
-  RESIZED_VM_SIZE="${RESIZED_VM_SIZE:-300}"
-else
-  RESIZED_VM_SIZE="${RESIZED_VM_SIZE:-100}"
-fi
 
 VM_TIMELABEL="${VM_TIMELABEL:-$(date '+%Y%m%d%H%M%S')}"
 TEST_EXECUTER_VM_NAME="${TEST_EXECUTER_VM_NAME:-ci-test-vm-${VM_TIMELABEL}}"
@@ -63,7 +57,7 @@ echo "Running in region: $OS_REGION_NAME"
 if [[ "${TESTS_FOR}" == "feature_tests"* || "${TESTS_FOR}" == "e2e_tests"* ]]
 then
     # Four node cluster
-    TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-200GB}"
+    TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-300GB}"
 else
     # user-data script for the 4C flavors in this region aren't correct for CentOS
     # so we run on a 8C machine to ensure the disk is properly resized


jenkins/scripts/integration_test.sh
```diff
@@ -44,7 +44,7 @@ if [[ "${TESTS_FOR}" == "feature_tests"* || "${TESTS_FOR}" == "e2e_tests"* ]]
 then
   RESIZED_VM_SIZE="${RESIZED_VM_SIZE:-300}"
 else
-  RESIZED_VM_SIZE="${RESIZED_VM_SIZE:-100}"
+  RESIZED_VM_SIZE="${RESIZED_VM_SIZE:-200}"
```
macaptain, line 47: I don't think we use RESIZED_VM_SIZE anywhere anymore, since we don't boot from volume. We can remove this block of code entirely.

jenkins/scripts/integration_test.sh
```diff
@@ -63,16 +63,16 @@ echo "Running in region: $OS_REGION_NAME"
 if [[ "${TESTS_FOR}" == "feature_tests"* || "${TESTS_FOR}" == "e2e_tests"* ]]
 then
     # Four node cluster
-    TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-200GB}"
+    TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-32GB-300GB}"
```
macaptain, line 66: The thing to check here is that these flavors actually exist. You can do `openstack flavor list` in both Kna1 and Fra1 regions in CityCloud to see if they do. If the flavors don't exist, making a VM with the specs using the GUI with the sliders will create the corresponding flavor in the CLI environment.

Rozzii: Yes they are presetn:
Kna1
| d27c5b84-ce03-45bd-a8c8-45094b4df750 | 8C-32GB-300GB              |  32768 |  300 |         0 |     8 | True      |
Fra1
| 3038f8e9-7a6a-4a9c-a6ff-0857f0e0f8c0 | 8C-32GB-300GB    |  32768 |  300 |         0 |     8 | True      |

jenkins/scripts/integration_test.sh
Rozzii: Thanks, I have removed it.

jenkins/scripts/integration_test.sh
```diff
@@ -63,16 +57,16 @@ echo "Running in region: $OS_REGION_NAME"
...
 else
     # user-data script for the 4C flavors in this region aren't correct for CentOS
     # so we run on a 8C machine to ensure the disk is properly resized
     if [[ "$OS_REGION_NAME" == "Fra1" ]] && [[ "$DISTRIBUTION" != "ubuntu" ]]
     then
-      TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-16GB-100GB}"
+      TEST_EXECUTER_FLAVOR="${TEST_EXECUTER_FLAVOR:-8C-16GB-200GB}"
```
furkatgofurov7, line 66: Do we need 200GB for 2 node cluster? They used to work pretty well with 100GB even before

Rozzii: I have no strong opinion on this, if you think 100 is enough I can leave this alone

furkatgofurov7: Me neither, but since we run most of the tests with 2 node cluster, we might need double of amount of storage for them compared to now. So, maybe let's increase the storage size for 4 node cluster only maybe?

Rozzii: I have removed the modification.

furkatgofurov7 (approved):
/approve
cc @macaptain

PR #290, opened 2021-12-15
Increase frequency of running go based CI jobs 

fmuyassarov:
Increase frequency of running go based CI jobs.
We should be more careful to when not to run go based
prow jobs. We recently faced a PR which didn't touch
`.go `file, so gofmt test didn't run but that change made
existing `.go` files to fail passing go fmt. As such, make it
more strict when the jobs should not run.

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -358,7 +358,7 @@ presubmits:
   - name: gofmt
     branches:
     - main
-    run_if_changed: '\.go$'
+    skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
     decorate: true
     spec:
       containers:
@@ -375,7 +375,7 @@ presubmits:
     branches:
     - release-0.5
     - release-0.4
-    run_if_changed: '\.go$'
+    skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
     decorate: true
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


prow/config/config.yaml
```diff
@@ -358,7 +358,7 @@ presubmits:
   - name: gofmt
     branches:
     - main
-    run_if_changed: '\.go$'
+    run_if_changed: '((\.go)|^hack/gofmt.sh)$'
```
macaptain, line 361: Nit: the `.` in `gofmt.sh` will match correctly. But probably you wanted to escape the `.` with `\.`.

fmuyassarov: thanks for the catch

prow/config/config.yaml
```diff
@@ -358,7 +358,7 @@ presubmits:
   - name: gofmt
     branches:
     - main
-    run_if_changed: '\.go$'
+    run_if_changed: '((\.go)|^hack/gofmt\.sh)$'
```
macaptain, line 361: Also, tiny thing, the parentheses around `\.go` aren't needed, I think.

macaptain (approved):

fmuyassarov:
Closing it, as it seems this might add more complexity to the prow configurations.
/close

kashifest (approved):

furkatgofurov7 (commented):
/lgtm

PR #291, opened 2021-12-15
Revert golang images to 1.16 for IPAM and CAPM3 main jobs

fmuyassarov:
Apparently CAPI hasn't switched yet to 1.17, so CAPM3
and IPAM should still use 1.16 too.


fmuyassarov:
/hold 
While branch migration to main is taking place

fmuyassarov:
/close
Since the issue was addressed by using CAPI beta release

PR #292, opened 2021-12-16
Improve BML cleaning

lentzi90:
- Power off BMHs before tearing down the source cluster
- Remove local container registry
- Prune containers, volumes and images

~~This depends on https://github.com/metal3-io/metal3-dev-env/pull/887~~

### DIFF
--- a/jenkins/scripts/bare_metal_lab/deploy-lab.yaml
+++ b/jenkins/scripts/bare_metal_lab/deploy-lab.yaml
@@ -11,10 +11,11 @@
         msg: Ensure that BML_ILO_USERNAME and BML_ILO_PASSWORD environment variables are set
       when: not bml_ilo_username or not bml_ilo_password
 
-    - name: Install dependencies for metal3-dev-env
+    - name: Install dependencies for metal3-dev-env and hpilo_boot
       apt:
         pkg:
           - make
+          - python3-hpilo
       become: true
 
     - name: Add passwordless sudo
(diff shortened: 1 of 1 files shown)


lentzi90:
After some more thinking I have changed this to simply power off the BMHs instead of deprovisioning. My reasoning is that deprovisioning will anyway happen at the end of a *successful* pipeline. If the pipeline is *not* successful, it is hard to say anything about what state the cluster and the machines are in. This means that there is a large risk that deprovisioning will fail anyway and could then block the pipeline. So instead of waiting on deprovisioning, simply power off.

Rozzii (approved):

jenkins/scripts/bare_metal_lab/deploy-lab.yaml
```diff
@@ -56,19 +57,43 @@
         version: master
       tags: git
 
+    - name: Power off BMHs
```
fmuyassarov, line 60: nit: this was a bit confusing for me, because it is not yet a BMH (k8s CRD) but actual server.

fmuyassarov (commented):
/approve
/hold 
@lentzi90 feel free to unhold, just didn't wanted the PR to get merged before you saw the comment.

jenkins/scripts/bare_metal_lab/deploy-lab.yaml
lentzi90: Right, I understand the confusion. I was thinking that the point is really to make sure that the BMHs from the previous pipeline are cleaned up. At this point the management cluster is still running with these BMHs registered, so in that sense they are BMHs.

PR #294, opened 2021-12-21
Set same variables for BML as for other pipelines

lentzi90:
Since the BML environment is set up in a different way than the other
integration tests, it didn't use the variables for local images,
container registry and similar. This PR makes it use the same
variables.

Set DOCKER_HUB_PROXY to use the Nordix Harbor instance. This variable is added in https://github.com/metal3-io/metal3-dev-env/pull/892 so before that is merged it won't do anything.

Make the BML pipeline prune all images (not just dangling), and clean
the metal3-dev-env repo between runs.


jenkins/scripts/bml_integration_test.sh
```diff
@@ -56,6 +55,7 @@ scp \
   -r \
   "${CI_DIR}/files/run_integration_tests.sh" \
   "${CI_DIR}/files/vars.sh" \
+  "${CI_DIR}/files/integration_test_env.sh" \
```
lentzi90, line 58: I really wanted to reuse `vars.sh` but I couldn't find a good solution. The issue is that in the other pipelines everything is set up through `run_integration_tests.sh` so it is enough to get the variables in the environment used there, but for BML we use the `deploy-lab.yaml` playbook so we need to have the same environment available there. Therefore I decided to use the `integration_test_env.sh` as a `config_{user}.sh` file with BML.

jenkins/scripts/bare_metal_lab/deploy-lab.yaml
```diff
@@ -2,14 +2,16 @@
...
     EPHEMERAL_CLUSTER: "minikube"
     EXTERNAL_VLAN_ID: 3
+    GITHUB_TOKEN: "{{ lookup('env', 'GITHUB_TOKEN') }}"
   tasks:
     - name: Check required env vars are set
       fail:
-        msg: Ensure that BML_ILO_USERNAME and BML_ILO_PASSWORD environment variables are set
-      when: not bml_ilo_username or not bml_ilo_password
+        msg: Ensure that BML_ILO_USERNAME, BML_ILO_PASSWORD and GITHUB_TOKEN environment variables are set
+      when: not bml_ilo_username or not bml_ilo_password or not github_token
```
fmuyassarov, line 14: it is weird why CI is failing here, can we perhaps try with 

```suggestion
        msg: Ensure that BML_ILO_USERNAME, BML_ILO_PASSWORD and GITHUB_TOKEN environment variables are set
      when: not (bml_ilo_username or bml_ilo_password or github_token)
```

fmuyassarov: disregard, this didn't help

fmuyassarov: I wonder if the variable is not getting exported properly from Jenkins

fmuyassarov:
/hold 
While branch migration to main is taking place

jenkins/scripts/bare_metal_lab/files/config.sh
```diff
@@ -1,168 +0,0 @@
...
-#
-# Choose whether the "baremetal" libvirt network will use IPv4, IPv6, or IPv4+IPv6.
-# This network is the primary network interface for the virtual bare metal hosts.
-#
-# Note that this only sets up the underlying network, and fully provisioning IPv6
-# kubernetes clusters is not yet automated.  If IPv6 is enabled, DHCPv6 will
-# be available to the virtual bare metal hosts.
-#
-# v4   -- IPv4 (default)
-# v6   -- IPv6
```
fmuyassarov, line 12: I think this file was there as for the reference of a doc to explain to user of BMH shortly what each variable is for. Do we really want to drop it?

jenkins/scripts/bare_metal_lab/README.md
```diff
@@ -1,17 +1,17 @@
...
+The Bare Metal Lab needs some special treatment compared to other pipelines since it does not use VMs for the target cluster.
+This is taken care of in the `deploy-lab.yaml` playbook.
 
 ## Ansible installation
 
 `sudo pip3 install ansible`
 
 ## Running the playbook
 
-* Set 'bm-m3-lab' to the IP address of the lab in /etc/hosts.
```
furkatgofurov7, line 11: why is this removed?

fmuyassarov (commented):
Export of env vars depending on the repo name (which you moved to a new file) should be exported before we run make targets of dev-env. Did I understand it right that those exports are now happening in the first ssh (i.e. `# Send Remote script to Executer`), and then in the next ssh (e`cho "Running the tests"`) we run the Makefile targets from dev-env ?

jenkins/scripts/bare_metal_lab/files/config.sh
```diff
@@ -1,168 +0,0 @@
...
-# Selecting "tilt" does not deploy a management cluster, it is left up to the
-# user
-
-# Secure Ironic deployment with TLS ("true" or "false")
-# export IRONIC_TLS_SETUP="true"
-
-# Set nodeDrainTimeout for controlplane and worker template, otherwise default value will be  '0s'. 
-#
-#export NODE_DRAIN_TIMEOUT=${NODE_DRAIN_TIMEOUT:-"0s"}
-
```
furkatgofurov7, line 168: I don't have a strong opinion on this, while this was used as a reference file to know what can be set or not, I am fine with removing it also, because the one in m3-dev-env keeps changing frequently and we always have to keep this file in sync with that.

jenkins/scripts/bare_metal_lab/deploy-lab.yaml
```diff
@@ -2,14 +2,16 @@
   vars:
     bml_ilo_username: "{{ lookup('env', 'BML_ILO_USERNAME') }}"
     bml_ilo_password: "{{ lookup('env', 'BML_ILO_PASSWORD') }}"
+    github_token: "{{ lookup('env', 'GITHUB_TOKEN') }}"
   environment:
     EPHEMERAL_CLUSTER: "minikube"
     EXTERNAL_VLAN_ID: 3
+    GITHUB_TOKEN: "{{ lookup('env', 'GITHUB_TOKEN') }}"
```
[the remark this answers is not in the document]
lentzi90: Good point! I added it while trying to figure out why it was failing and forgot to remove it. It should work also without this so I'll remove it and rerun the tests :slightly_smiling_face:

jenkins/scripts/bare_metal_lab/files/config.sh
lentzi90: I thought this was just a duplicate of [the one we have in metal3-dev-env](https://github.com/metal3-io/metal3-dev-env/blob/master/config_example.sh). Do we need both?

jenkins/scripts/bare_metal_lab/README.md
lentzi90: It does not seem it is used/needed. I'm guessing there used to be a `hosts` file that had `bm-m3-lab` in it but this is no longer the case. The playbook is hard coded to run on `localhost` and the pipeline is not using any inventory, just `ansible-playbook <path-to-playbook>`.

jenkins/scripts/bare_metal_lab/files/config.sh
fmuyassarov: Ah I see, didn't notice it. Yeah, I would be +1 for dropping it, not sure why we added

jenkins/scripts/bare_metal_lab/files/config.sh
lentzi90: Yes I was thinking that it was confusing to have it here also, and since it is not really used other than for docs I think it makes sense to remove it from this repo and keep it in metal3-dev-env.

furkatgofurov7: +1

furkatgofurov7 (commented):
/lgtm

PR #298, opened 2021-12-23
📖 Update jenkins readme

furkatgofurov7:


kashifest (approved):

jenkins/README.md
```diff
@@ -33,28 +33,13 @@ The job result will be posted as a comment.
 * **/test-v1a4-centos-integration** run integration tests for v1alpha4 on
    CentOS
 
-It is also possible to prevent any job run by adding **/skip-test** in the PR
```
furkatgofurov7, line 36: removed this, because this is not supported

jenkins/README.md
```diff
@@ -33,28 +33,13 @@ The job result will be posted as a comment.
...
-description.
-
-If the author is not in the whitelist but should be trusted then by adding a
-comment **add to whitelist** on the PR, the author will then be able to run the
-jobs on its own.
```
furkatgofurov7, line 41: this also, since it is already mentioned a bit earlier

jenkins/README.md
```diff
@@ -33,28 +33,13 @@ The job result will be posted as a comment.
...
-
 ## Cloud Resources cleanup
 
 There is a Jenkins [master job](https://jenkins.nordix.org/view/Airship/job/airship_master_integration_tests_cleanup/)
 that cleans up all the leftover VMs from
 [CityCloud](https://www.citycloud.com/) every 6 hours which has failed to be
 deleted at the end of v1alphaX integration test.
 
-## "Can one of the admins verify this patch?"
```
furkatgofurov7, line 50: same, bot does not add such comment

furkatgofurov7 (commented):
To clarify some points:

jenkins/README.md
```diff
@@ -14,7 +14,7 @@ publicly visible will get admin rights on the CI jobs. This means :
 * They can start the jobs on their PR directly
 * They can start the jobs for PR of authors that are not in the organization
 * They can add authors to whitelist so that the authors can start jobs on any
-   further PR on their own, by commenting **add to whitelist** on the PR
+   further PR on their own, by commenting **/ok-to-test** on the PR
```
fmuyassarov, line 17: This doesn't seem correct to me. Adding /ok-to-test doesn't allow the person to start jobs on the further PRs

jenkins/README.md
```diff
@@ -33,28 +33,13 @@ The job result will be posted as a comment.
...
 deleted at the end of v1alphaX integration test.
 
-## "Can one of the admins verify this patch?"
-
-For all the PRs from authors that are not whitelisted, the bot will add a
-comment "*Can one of the admins verify this patch?*". This means that the author
-is not in the whitelist and that someone from the metal3-io organization should
-review the PR and run the tests (with */test-integration*) or add the author to
-whitelist if trusted.
-
```
fmuyassarov, line 57: instead of removing, it can be updated to

```
. Thanks for your PR.

I'm waiting for a metal3-io member to verify that this patch is reasonable to test. If it is, they should reply with /ok-to-test on its own line. Until that is done, I will not automatically test new commits in this PR, but the usual testing commands by org members will still work. Regular contributors should join the org to skip this step.

Once the patch is verified, the new status will be reflected by the ok-to-test label.

I understand the commands that are listed here.
```
example: https://github.com/metal3-io/baremetal-operator/pull/964#issuecomment-911958716

fmuyassarov (changes requested):

jenkins/README.md
furkatgofurov7: Sure, clarified it a bit more

jenkins/README.md
furkatgofurov7: Updated

fmuyassarov (commented):
/lgtm

PR #299, opened 2021-12-23
Disable concurrent builds in BML

lentzi90:
The builds in BML always use the same machine so we cannot allow multiple builds to run at the same time.

### DIFF
--- a/jenkins/scripts/bare_metal_lab/README.md
+++ b/jenkins/scripts/bare_metal_lab/README.md
@@ -15,3 +15,16 @@ This repo is created to setup the baremetal lab for metal3.
 Then:
 
 `ansible-playbook ./deploy-lab.yaml -i ./hosts -u <user> --ask-become-pass`
+
+## Running tests for pull requests on Github
+
+You can trigger builds to run in the bare metal lab by adding the following line as a comment on the PR:
+
+```
+/test-integration-bml-centos
+```
+
+**Note:** Concurrent builds are disabled for the BML, since they would run on the same host and interfere with each other.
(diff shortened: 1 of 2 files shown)


lentzi90:
Tested by adding the `disableConcurrentBuilds()` option on another PR, triggering the pipeline there, then trigger this one.
The build is correctly triggered and shows as Pending while the other build is running.
pending-build

furkatgofurov7 (commented):
Small suggestion inline, otherwise lgtm

jenkins/scripts/bare_metal_lab/README.md
```diff
@@ -15,3 +15,16 @@ This repo is created to setup the baremetal lab for metal3.
...
+
+You can trigger builds to run in the bare metal lab by adding the following line as a comment on the PR:
+
+```
+/test-integration-bml-centos
+```
+
+**Note:** Concurrent builds are disabled for the BML, since they would run on the same host and interfere with each other.
+This means that if there is already one build job running in the BML, a new one will not start before the first has finished.
+Github will show that the build is pending, but there is *no link* to Jenkins at this time.
```
furkatgofurov7, line 29:
```suggestion
Github won't show the usual *Details* link for this specific run but build status can be checked from the 
 [Jenkins dashboard](https://jenkins.nordix.org/view/Airship/job/airship_master_bml_integration_tests_centos/)
 where the build will be scheduled and stay in pending at this time.
```

fmuyassarov (commented):
/approve

furkatgofurov7 (approved):
/lgtm

PR #300, opened 2021-12-31
Create/delete floating IP in Fra1

lentzi90:
This is to avoid collisions when multiple jobs would happen to select
the same IP. If each job creates its own IP we get no collisions.

### DIFF
--- a/jenkins/scripts/integration_delete.sh
+++ b/jenkins/scripts/integration_delete.sh
@@ -3,7 +3,7 @@
 set -eu
 
 # Description:
-# Runs in every single main job and in jobs triggered within the PR in metal3 repos. 
+# Runs in every single main job and in jobs triggered within the PR in metal3 repos.
 # Consumed by integration_tests.pipeline and cleans the executer vm/volume and port after
 # integration tests.
 #   Requires:
@@ -18,6 +18,7 @@ CI_DIR="$(dirname "$(readlink -f "${0}")")"
 source "${CI_DIR}/utils.sh"
 
 TEST_EXECUTER_PORT_NAME="${TEST_EXECUTER_PORT_NAME:-${TEST_EXECUTER_VM_NAME}-int-port}"
+TEST_EXECUTER_FIP_TAG="${TEST_EXECUTER_FIP_TAG:-${TEST_EXECUTER_VM_NAME}-floating-ip}"
 
 # Run feature tests, e2e tests, main, master and release* tests in the Frankfurt region
 if [[ "${TESTS_FOR}" == "feature_tests"* ]] || [[ "${TESTS_FOR}" == "e2e_tests"* ]] || \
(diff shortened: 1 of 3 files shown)


jenkins/scripts/integration_delete.sh
```diff
@@ -28,6 +29,17 @@ then
 fi
 echo "Running in region: $OS_REGION_NAME"
 
+if [[ "$OS_REGION_NAME" != "Kna1" ]]
+then
+  # Find executer floating ip
+  TEST_EXECUTER_FIP_ID="$(openstack floating ip list --tags "${TEST_EXECUTER_FIP_TAG}" -f json | jq -r '.[].ID')"
```
lentzi90, line 35: Theoretically we could get more than one IP here. Should we loop over them and delete instead of assuming we only get one?

macaptain: You could also pipe to `xargs`.

jenkins/scripts/integration_test.sh
```diff
@@ -88,10 +89,11 @@ TEST_EXECUTER_IP="$(openstack port show -f json "${TEST_EXECUTER_PORT_NAME}" \
...
 then
-  # Fetch Free floating IP
-  # TODO: To avoid jobs taking the same floating IP, instead jobs should create / delete / cleanup
-  # their own floating IPs.
-  FLOATING_IP="$(openstack floating ip list --status DOWN -c "Floating IP Address" -f value | shuf -n 1 )"
+  # Create floating IP
+  FLOATING_IP="$(openstack floating ip create -f json \
+    --tag "${TEST_EXECUTER_FIP_TAG}" \
+    --subnet "$(get_subnet_name "${CI_EXT_NET}")" \
+    "${CI_EXT_NET}" | jq -r '.name')"
```
macaptain, line 96: `-f value -c name` would get the name of the floating IP without shelling out to `jq`.

macaptain (approved):
Great! Then we can remove the IP addresses that are currently created and 'pooled'. 

/lgtm

Two minor suggestions in review.

It would be nice to test this, but running the integration tests from this branch will run on Kna1. A feature test would run on Fra1, but would take some time. I wonder if we should add a Prow command that lets us specifically test on Fra1. :thinking: In any case, I would be OK with merging and running a `main` build to verify the change, since it's a small patch.

lentzi90:
Good thinking! I'll simply trigger a feature test!
/test-features

jenkins/scripts/integration_delete.sh
lentzi90: Today I learned about `xargs` :smile: 
Excellent idea!

fmuyassarov:
> Great! Then we can remove the IP addresses that are currently created and 'pooled'.
> 
> /lgtm
> 
> Two minor suggestions in review.
> 
> It would be nice to test this, but running the integration tests from this branch will run on Kna1. A feature test would run on Fra1, but would take some time. I wonder if we should add a Prow command that lets us specifically test on Fra1. thinking In any case, I would be OK with merging and running a `main` build to verify the change, since it's a small patch.

@macaptain unfortunately we can't pass region from Prow, because integration tests are based on Jenkins.

fmuyassarov (approved):

PR #301, opened 2022-01-04
Allow keeping tester VM for 24 hours

fmuyassarov:
Allow keeping tester virtual machine for 24 hours for debugging
purposes. Depends on https://gerrit.nordix.org/c/infra/cicd/+/11894

Trigger phrases:
```
keep-test-v1a5-centos-integration
keep-test-v1a5-integration
keep-test-v1b1-centos-integration
keep-test-v1b1-integration
```


jenkins/jobs/integration_tests.pipeline
```diff
@@ -60,6 +60,7 @@ pipeline {
     NUM_NODES="${NUM_NODES}"
     TESTS_FOR="${TESTS_FOR}"
     PROMETHEUS_MONITORING="${PROMETHEUS_MONITORING}"
+    TTL="{TTL}"
```
Rozzii, line 63: "$" is missing otherwise it looks good,

Rozzii (changes requested):

jenkins/jobs/integration_tests.pipeline
```diff
@@ -121,7 +121,13 @@ pipeline {
     success {
       withCredentials([usernamePassword(credentialsId: 'airshipci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')]) {
         withCredentials([sshUserPrivateKey(credentialsId: 'airshipci_city_cloud_ssh_keypair', keyFileVariable: 'AIRSHIP_CI_USER_KEY')]){
-          sh "./jenkins/scripts/integration_test_clean.sh"
+          script {
+            if ("{TTL}" != "") {
+              echo "Going to sleep for "{TTL}" hours"
+              sleep(time:"{TTL}",unit:"HOURS")
```
macaptain, line 127: Interesting. I'm not sure I like the idea of very long running Jenkins builds to keep resources around. A long running Jenkins job will continue to consume an executor on the agent, even if it is sleeping.

What if instead, we have an option to skip the cleanup? The VM will eventually be cleaned up by the integration tests cleanup job:
https://jenkins.nordix.org/view/Airship/job/airship_master_integration_tests_cleanup

Currently looks like the machines are cleaned up after 6 hours: 
https://github.com/metal3-io/project-infra/blob/cc846fe9835d1373a7eb0954c17b24eaf0d4ec69/jenkins/scripts/integration_clean.sh#L23

Potentially we could be less aggressive here. Or alternatively, you could tag the VM and use `--not-tags` to skip deletion. (We'd also have to do the same for the ports.)

Rozzii: I would support the tagging of VMs that were started with the "keep 24h" jobs and the modification of the cleanup script.
We could also limit the number of VMs kept alive with the "keep" commands if we limit the number of active "keep" builds to 1 per job.

fmuyassarov: This sounds a better approach but a bit confused. If we are skipping the clean up and deletion, who will do the cleaning later on (run make clean)? For the VM deletion we have a script that runs periodically and we can make it so that after 24 hours it catches the "keep 24h" VM and deletes it. However, before that we need to run make clean as well, which is part of https://github.com/metal3-io/project-infra/blob/cc846fe9835d1373a7eb0954c17b24eaf0d4ec69/jenkins/jobs/integration_tests.pipeline#L124

furkatgofurov7 (commented):
Some questions regarding the scope of this addition: how about if we want to keep VM not only for integration tests but for e2e or feature tests too? The reason is from my experience mostly those long-running jobs need keeping VM alive.

fmuyassarov:
> Some questions regarding the scope of this addition: how about if we want to keep VM not only for integration tests but for e2e or feature tests too? The reason is from my experience mostly those long-running jobs need keeping VM alive.

The plan is to keep the VM for ~24h and for now I've done it for only integration tests in the Jenkins Job builder. We can extend the list to e2e too if there is a need, I just wasn't sure if we needed it.

furkatgofurov7:
> We can extend the list to e2e too if there is a need, I just wasn't sure if we needed it.

I do vote at least extending it to add e2e tests, even though we have a [different hack](https://github.com/metal3-io/cluster-api-provider-metal3/pull/343/files#diff-fe6e748ccfdc1ed1783825a0c4e64208466f536d94caedb88ae0930be0581b98R92) to keep VM infinitely, most of the people who are active in e2e test development would be interested to have it IMO.

fmuyassarov:
> 

isn't that patch for pod restart timeout ?

jenkins/scripts/integration_clean.sh
```diff
@@ -22,10 +22,17 @@ function cleanup() {
     startswith("ci-test-vm-")) | select((.Name | ltrimstr("ci-test-vm-") |
     split("-") | .[0] | strptime("%Y%m%d%H%M%S") | mktime) < (now - 21600))
     | .ID ')
+
+  DEBUGGING_VM_LIST=$(openstack server list -f json | jq -r '.[] | select(.Name |
+  startswith("ci-test-vm-")) | select((.Name | ltrimstr("ci-test-keep-vm-") |
+  split("-") | .[0] | strptime("%Y%m%d%H%M%S") | mktime) < (now - 86400))
+  | .ID ')
   
+  FINAL_LIST=$(IFS=$'\n'; echo "${VM_LIST[*]}" "${DEBUGGING_VM_LIST[*]}")
```
fmuyassarov, line 31: I should admit I am bad with naming of vars. If you have better option please let me know

jenkins/scripts/integration_clean.sh
```diff
@@ -22,10 +22,17 @@ function cleanup() {
     startswith("ci-test-vm-")) | select((.Name | ltrimstr("ci-test-vm-") |
     split("-") | .[0] | strptime("%Y%m%d%H%M%S") | mktime) < (now - 21600))
     | .ID ')
+
+  DEBUGGING_VM_LIST=$(openstack server list -f json | jq -r '.[] | select(.Name |
+  startswith("ci-test-vm-")) | select((.Name | ltrimstr("ci-test-keep-vm-") |
```
macaptain, line 27: I expect a VM named 'ci-test-keep-vm-' won't match the selection of `startswith("ci-test-vm-")`.

jenkins/scripts/integration_clean.sh
```diff
@@ -22,10 +22,17 @@ function cleanup() {
...
+  split("-") | .[0] | strptime("%Y%m%d%H%M%S") | mktime) < (now - 86400))
```
macaptain, line 28: Could we replace the magic numbers 21600 and 86400 with variables that indicate the length of time in a human readable way?

jenkins/jobs/integration_tests.pipeline
```diff
@@ -121,14 +121,24 @@ pipeline {
     success {
       withCredentials([usernamePassword(credentialsId: 'airshipci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')]) {
         withCredentials([sshUserPrivateKey(credentialsId: 'airshipci_city_cloud_ssh_keypair', keyFileVariable: 'AIRSHIP_CI_USER_KEY')]){
-          sh "./jenkins/scripts/integration_test_clean.sh"
+          script {
+            if ("{SKIP_DELETION}" == "true") {
+              echo "Skipping environment clean up"
+            }
+            sh "./jenkins/scripts/integration_test_clean.sh"
```
lentzi90, line 128: I think there should be an `else` here, right?

jenkins/jobs/integration_tests.pipeline
```diff
@@ -121,14 +121,24 @@ pipeline {
...
     }
     cleanup {
       withCredentials([usernamePassword(credentialsId: 'airshipci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')]) {
         withCredentials([sshUserPrivateKey(credentialsId: 'airshipci_city_cloud_ssh_keypair', keyFileVariable: 'AIRSHIP_CI_USER_KEY')]){
-          sh "./jenkins/scripts/integration_delete.sh"
+          script {
+            if ("{SKIP_DELETION}" == "true") {
+            echo "Skipping VM deletion"
+            }
+            sh "./jenkins/scripts/integration_delete.sh"
```
lentzi90, line 140: Here also `else`.

jenkins/scripts/integration_clean.sh
fmuyassarov: I've added a comment on that

jenkins/scripts/integration_test_clean.sh
```diff
@@ -20,6 +20,8 @@ DISTRIBUTION="${DISTRIBUTION:-ubuntu}"
 # shellcheck disable=SC1090
 source "${CI_DIR}/utils.sh"
 
+
+
```
furkatgofurov7, line 24: nit: double space added

jenkins/scripts/integration_clean.sh
```diff
@@ -37,8 +46,15 @@ function cleanup() {
     startswith("ci-test-vm-")) | select((.Name | ltrimstr("ci-test-vm-") |
     rtrimstr("-int-port") | split("-") | .[0] | strptime("%Y%m%d%H%M%S") |
     mktime) < (now - 21600)) | .ID ')
+
+  DEBUGGING_VM_PORT_LIST=$(openstack port list -f json | jq -r '.[] | select(.Name |
```
furkatgofurov7, line 50: Would be great to have comments for both port types too re: hours in seconds.

furkatgofurov7 (commented):
Overall looks good, some comments inline. Also, we should hold this until jjb is in place?

furkatgofurov7:
I would strongly suggest documenting all test trigger phrases introduced in this PR for future reference.

jenkins/scripts/integration_test.sh
```diff
@@ -37,11 +37,8 @@ CAPM3_VERSION="${CAPM3_VERSION:-v1beta1}"
 NUM_NODES="${NUM_NODES:-2}"
 TESTS_FOR="${TESTS_FOR:-integration_test}"
 BARE_METAL_LAB=false
-
 DEFAULT_HOSTS_MEMORY="${DEFAULT_HOSTS_MEMORY:-4096}"
-
-VM_TIMELABEL="${VM_TIMELABEL:-$(date '+%Y%m%d%H%M%S')}"
```
furkatgofurov7, line 43: Is there any specific reason we do not want the time tag to be appended to the executer VM name?

jenkins/README.md
```diff
@@ -33,6 +33,20 @@ The job result will be posted as a comment.
 * **/test-v1a4-centos-integration** run integration tests for v1alpha4 on
    CentOS
 
+Usually after integration test part is completed, Jenkins executes another script
```
furkatgofurov7, line 36: Thanks, this helps a lot :+1: 
```suggestion
Usually, after the integration test part is completed, Jenkins executes another script
```

jenkins/README.md
```diff
@@ -33,6 +33,20 @@ The job result will be posted as a comment.
...
+to clean up the environment first and then deletes the VM. However, sometimes
+it may be desirable to keep the VM for debugging purposes. To avoid clean up
+and deletion operations, there are seperate triggers phrases as below:
```
furkatgofurov7, line 39:
```suggestion
and deletion operations, there are separate triggers phrases as below:
```

jenkins/README.md
```diff
@@ -33,6 +33,20 @@ The job result will be posted as a comment.
...
+
+* **/keep-test-v1b1-integration** run integration tests for v1beta1 on Ubuntu
+* **/keep-test-v1b1-centos-integration** run integration tests for v1beta1 on
+   CentOS
+* **/keep-test-v1a5-integration** run integration tests for v1alpha5 on Ubuntu
+* **/keep-test-v1a5-centos-integration** run integration tests for v1alpha5 on
```
furkatgofurov7, line 45:
```suggestion
* **/keep-test-v1a5-centos-integration** run integration tests for v1alpha5 on CentOS
```

furkatgofurov7 (commented):
Some nits and questions are inline, overall this is shaping nicely. 
One more general question: Kept without being deleted VMs won't have floating IPs and not a personal ssh key in it, meaning the person should ssh into it using the, let's say <any VM accessible> ssh keys?

jenkins/scripts/integration_test.sh
fmuyassarov: yes, we were relayin on that in case we need to give a name to VM within this script. But in reality we never do it, because naming of the VM takes place earlier than we reach here, in the pipeline. And in the integration test pipeline we have a script that does the same, but in different env var `VM_KEY`

`VM_KEY = (1..4).collect { ('a'..'z').join("")[ rand.nextInt( 26 ) ] }.join("")`

fmuyassarov:
> Some nits and questions are inline, overall this is shaping nicely. One more general question: Kept without being deleted VMs won't have floating IPs and not a personal ssh key in it, meaning the person should ssh into it using the, let's say "any VM accessible" ssh keys?

For now we would like to allow only CI team to have the access due to the way we set ssh keys on the VM. But we might think about possible ways to extend the access to the community in the future.

macaptain:
I don't think the test-v1a5-integration test should be failing?
/hold

furkatgofurov7:
Looks good to me. Triggering all tests to make sure all trigger phrases working as expected.
/approve
/test-v1b1-integration
/keep-test-v1b1-integration
/keep-test-v1a5-centos-integration
/keep-test-v1b1-centos-integration

PR #310, opened 2022-01-14
Avoid using empty variables

lentzi90:
The alternative would be to always set some default value for these release variables. However, it is quite nice to get the default from metal3-dev-env, which is always checking what the latest release is.


kashifest (approved):

furkatgofurov7 (commented):
/lgtm

PR #311, opened 2022-01-14
Fix unbound variable use logic

lentzi90:
I made a logic error in the last PR. It should be setting these variables only *if they have some value*. If they are unset or empty string we should ignore them to allow default values to be used.

/hold
Don't merge this until [this](https://gerrit.nordix.org/c/infra/cicd/+/12038) is in so we can properly test it.


furkatgofurov7 (approved):
/approve

/cc @fmuyassarov @macaptain

Issue #322, opened 2022-01-27
Add yaml linter job to CI


furkatgofurov7:
As of now, in project-infra we only have `check-prow-config ` tests as default and mandatory, and other integration test triggers can also be triggered per need basis. However, we need to look forward to adding a new YAML linter job to CI that checks and catches unexpected format, blank space etc in YAML files being changed or added. 
 The need for this is because there were multiple cases up until now, where we had inappropriate YAML files after making changes (extra blank space addition mostly) to prow config files and that merged unnoticed, resulting in the unformatted and hard to read k8s resource YAML definitions being created in the prow cluster.

PR #327, opened 2022-01-30
Add linter CI for YAML files

fmuyassarov:
Currently we don't lint YAML files. As a result, sometimes we see
issues with some YAML code being merged and the issue is identified
later on. Example, trailing spaces which might breake the
configuration file of Prow.

This patch adds yamllinter based GitHub action so that we verify
correctness of YAML files on every pull request submission, before
landing on the main branch. As a start, trailing space, key-duplicate
and truthy rules are enabled, and the list can be extended based on
the need.
Fixes: https://github.com/metal3-io/project-infra/issues/322

### DIFF
--- a/.github/workflows/yamllint.yaml
+++ b/.github/workflows/yamllint.yaml
@@ -1,0 +1,16 @@
+---
+name: yamllint
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: yaml-lint
+        uses: ibiqlik/action-yamllint@v3
+        with:
+          config_file: .yamllint.yaml
--- a/.yamllint.yaml
+++ b/.yamllint.yaml
@@ -1,0 +1,10 @@
+yaml-files:
+  - '*.yaml'
+  - '*.yml'
+  - '.yamllint'
+
+rules:
+  trailing-spaces: enable
+  key-duplicates: enable
+  truthy:
+    allowed-values: ['true', 'false']
--- a/.yamllint.yaml
+++ b/.yamllint.yaml
@@ -7,4 +7,4 @@ rules:
   trailing-spaces: enable
   key-duplicates: enable
   truthy:
-    allowed-values: ['true', 'false']
+    allowed-values: ['true', 'false', 'yes', 'no', 'on']
--- a/jenkins/scripts/bare_metal_lab/deploy-lab.yaml
+++ b/jenkins/scripts/bare_metal_lab/deploy-lab.yaml
@@ -8,8 +8,8 @@
     EXTERNAL_VLAN_ID: 3
   tasks:
     - name: set_fact
-      set_fact: 
-        node_ips: 
+      set_fact:
+        node_ips:
           - "192.168.1.12"
           - "192.168.1.15"
         serial_log_location: "/tmp/BMLlog"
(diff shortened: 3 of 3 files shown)


.github/workflows/yamllint.yaml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install yamllint
+        run: pip install yamllint
```
macaptain, line 19: Should we use an image which has yamllint already installed? (e.g. https://hub.docker.com/r/cytopia/yamllint). That way we won't have to install it every time (once the image has been pulled into Prow).

fmuyassarov: My initial plan was the same and went this way, because I could not find an official image. But perhaps it is fine to use the one referenced.

fmuyassarov: Just heads up, that this is running on GitHub infra.  @macaptain  Do you think it is ok to use https://hub.docker.com/r/cytopia/yamllint ?

furkatgofurov7: This looks interesting, we can leave it as is as I found [this](https://github.com/lsst/pipe_base/blob/main/.github/workflows/yamllint.yaml#L18-L19) where it is installed the same way or maybe we could use this [one](https://github.com/frenck/action-yamllint#usage)?

furkatgofurov7 (commented):
Looks good, is there any way to test this patch before merging?

fmuyassarov:
> Looks good, is there any way to test this patch before merging?

yep, I have tested it on my work and works fine. Example: https://github.com/fmuyassarov/project-infra/pull/1

.github/workflows/yamllint.yaml
fmuyassarov: nice, I think using the GA image is the best, I will test it

fmuyassarov: I found `ibiqlik/action-yamllint@v3` image to be more stable and reliable.

Rozzii:
@fmuyassarov  Have you tested this on metal3-dev-env Ansible files? did it work correctly?

fmuyassarov:
> @fmuyassarov Have you tested this on metal3-dev-env Ansible files? did it work correctly?

yes, here is an example https://github.com/fmuyassarov/project-infra/pull/2. Btw, this is not meant to test dev-env scripts (YAML files) for only Prow yaml files in project-infra. Perhaps we can extend it later on, but that's out of scope for now.

.github/workflows/yamllint.yaml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
```
fmuyassarov, line 16: I believe this is leftover. I will remove it now

lentzi90 (commented):
Great addition! I would suggest enabling [key-duplicates](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.key_duplicates) and [truthy](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy) also as I think they can prevent some mistakes that are easy to miss otherwise.

.github/workflows/yamllint.yaml
fmuyassarov: tested and removed

namnx228 (approved):
LGTM

fmuyassarov:
> Great addition! I would suggest enabling [key-duplicates](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.key_duplicates) and [truthy](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy) also as I think they can prevent some mistakes that are easy to miss otherwise.

Yep, good suggestion. I will add it now
/hold

fmuyassarov:
@lentzi90 added those two rules.
/hold cancel

lentzi90 (commented):
/lgtm

furkatgofurov7 (approved):
Let's try it!

/approve

furkatgofurov7:
Prow is not behaving properly, trying it again.

/lgtm

kashifest (commented):
/lgtm

PR #307, opened 2022-01-11
Allow adding triage labels on issues across Metal3.io

fmuyassarov:
Allow setting needs-triage label on GitHub issues. This will tag
issues so that community is aware and takes some actions on them
during next triage meeting. Once action is taken, triage/foo should
be set indicating the triage status of the issue. Possible options

triage/unresolved, triage/not-reproducible, triage/duplicate,
triage/needs-information, triage/accepted.

PR #331, opened 2022-01-31
Prow: Add missing require-matching-label plugin to plugins list

furkatgofurov7:
We have enabled needs/triage label earlier in https://github.com/metal3-io/project-infra/pull/307 but that commit misses adding `require-matching-label` plugin to the list of plugins used across Metal3.io repos. This adds that plugin to the list.

### DIFF
--- a/prow/config/plugins.yaml
+++ b/prow/config/plugins.yaml
@@ -26,6 +26,7 @@ plugins:
     - verify-owners
     - wip
     - yuks
+    - require-matching-label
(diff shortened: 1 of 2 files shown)


furkatgofurov7:
Tested in issues:

- https://github.com/metal3-io/project-infra/issues/330
- https://github.com/metal3-io/metal3-dev-env/issues/925

kashifest (approved):

PR #335, opened 2022-02-02
Remove TARGET_NODE_OS variable

### LINKED ISSUE #334: Merge parameters DISTRIBUTION and TARGET_NODE_OS
These parameters are always set to either `ubuntu` or `centos`, but with the difference that `TARGET_NODE_OS` has it capitalized and `DISTRIBUTION` does not.

```
TARGET_NODE_OS=Ubuntu
DISTRIBUTION=ubuntu
```

We never mix them so that we have Ubuntu and centos, which means that we could drop one of them. This would simplify the pipelines and reduce the number of variables. The suggestion is to drop `DISTRIBUTION` and use `TARGET_NODE_OS` instead. Note that [TARGET_NODE_OS is translated to IMAGE_OS](https://github.com/metal3-io/project-infra/blob/ddba1f56e09306a09c5491d2e45eb001408b3dcd/jenkins/jobs/integration_tests.pipeline#L67) in the pipeline env.

lentzi90:
The TARGET_NODE_OS variable is always the same as DISTRIBUTION, except
that it is capitalized. This commit removes TARGET_NODE_OS and sets the
relevant values from DISTRIBUTION instead so that the number of
variables in reduced. This will help simplify the jjb templates.

Fixes #334


jenkins/scripts/bml_integration_test.sh
```diff
@@ -19,6 +19,13 @@ CI_DIR="$(dirname "$(readlink -f "${0}")")"
 # shellcheck disable=SC1090
 source "${CI_DIR}/utils.sh"
 
+if [ "${DISTRIBUTION}" == "ubuntu" ]
+then
+  IMAGE_OS="Ubuntu"
+else
+  IMAGE_OS="Centos"
+fi
```
lentzi90, line 27: I guess we could hard code it to "Centos" for the BML but this keeps things similar to the other pipelines.

fmuyassarov:
This looks good. And perhaps can be improved further, like `IMAGE_OS` and `DISTRIBUTION` can be merged into one for the CI only. For example, we could drop `DISTRIBUTION` var, and only use `IMAGE_OS` within JJB, which will be used for both source and target cluster nodes.
/approve

lentzi90:
> This looks good. And perhaps can be improved further, like `IMAGE_OS` and `DISTRIBUTION` can be merged into one for the CI only. For example, we could drop `DISTRIBUTION` var, and only use `IMAGE_OS` within JJB, which will be used for both source and target cluster nodes. /approve

Yes indeed I will do a PR for this in JJB! :slightly_smiling_face:

jenkins/scripts/bml_integration_test.sh
macaptain: One day Ubuntu might work in the BML.

PR #340, opened 2022-02-09
:seedling: Remove russellb as active approver

Rozzii:
Russell is no longer focusing on this project.


furkatgofurov7 (commented):
Thanks @russellb for all your contributions!

/lgtm
/retitle :seedling: Remove russellb as active approver

PR #341, opened 2022-02-09
Use metal3ci user for CI runs

macaptain:
This will swap to use the metal3ci user in the image instead of the airshipci user. After this, other pipelines will be updated and then the old airshipci user can be removed from use.


furkatgofurov7:
Looks like source branch is way more behind the main where we had volume usage and tests are failing.

Rozzii (approved):
LGTM

smoshiur1237 (approved):
/lgtm

kashifest (approved):

PR #355, opened 2022-02-28
Retry openstack server creation if unhealthy

lentzi90:
This is an attempt to work around the "no space" issue that is caused by a failed filesystem resize.

You can see the retry working in this run: https://jenkins.nordix.org/job/metal3_project_infra_v1a5_integration_test_centos/38/console
Unfortunately it failed later while deprovisioning, but everything else worked. It is quite hard to reproduce the issue so I have not tried to get another successful run with the retry.
I used a temporary commit for running in Fra1 to reproduce it more easily. This is now removed.

### DIFF
--- a/jenkins/scripts/utils.sh
+++ b/jenkins/scripts/utils.sh
@@ -45,3 +45,51 @@ wait_for_ssh() {
 
   echo "SSH connection to host[${SERVER}] is up."
 }
+
+# Description:
+# Check that the VM is in working condition, e.g. that the file-system is
+# resized properly.
+#
+# Usage:
+#   vm_healthy <ssh_user> <ssh_key_path> <server>
+vm_healthy() {
+  local USER KEY SERVER
+
+  USER="${1:?}"
+  KEY="${2:?}"
+  SERVER="${3:?}"
+
+  # CentOS has a cloud-init error all the time so we cannot rely on cloud-init
+  # to check if it is healthy or not.
+  # TODO: Fix cloud-init error on centos so we can remove this special check
+  if [ "${IMAGE_OS}" == "centos" ]; then
+    lsblk_out=$(ssh -o ConnectTimeout=2 -o StrictHostKeyChecking=no \
+      -o UserKnownHostsFile=/dev/null -i "${KEY}" \
+      "${USER}"@"${SERVER}" lsblk --json)
+    disk_data="$(echo "${lsblk_out}" | jq -r '.blockdevices[] | select(.name == "vda") | {disk_size: .size, part_size: .children[0].size}')"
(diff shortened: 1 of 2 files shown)


jenkins/scripts/utils.sh
```diff
@@ -45,3 +45,51 @@ wait_for_ssh() {
...
+  local USER KEY SERVER
+
+  USER="${1:?}"
+  KEY="${2:?}"
+  SERVER="${3:?}"
+
+  # CentOS has a cloud-init error all the time so we cannot rely on cloud-init
+  # to check if it is healthy or not.
+  # TODO: Fix cloud-init error on centos so we can remove this special check
+  if [ "${IMAGE_OS}" == "centos" ]; then
```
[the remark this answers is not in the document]
fmuyassarov: well since it is temporary, perhaps no need because I expect we will fully drop that func anyways once we have the actual fix. So, I'm fine with both ways.
/approve

jenkins/scripts/utils.sh
```diff
@@ -45,3 +45,51 @@ wait_for_ssh() {
...
+      echo "Filesystem resizing failed!"
+      echo "Disk size: ${disk_size}"
+      echo "Partition size: ${part_size}"
+      return 1
+    fi
+  fi
+
+  cloud_init_status=$(ssh -o ConnectTimeout=2 -o StrictHostKeyChecking=no \
+    -o UserKnownHostsFile=/dev/null -i "${KEY}" \
+    "${USER}"@"${SERVER}" cloud-init status --long --wait)
```
Xenwar, line 86: lgtm, if this works without `sudo`

jenkins/scripts/utils.sh
```diff
@@ -45,3 +45,51 @@ wait_for_ssh() {
...
+  if echo "${cloud_init_status}" | grep "error"; then
+    echo "There was a cloud-init error:"
+    echo "${cloud_init_status}"
+    return 1
+  else
+    echo "Server is healthy"
```
furkatgofurov7, line 92: This could be more specific, considering we do echo same message a bit earlier when checking disk resize part on line 74, but here we do cloud init status check so I would be happy to see something like, i.e:
   `echo "There was no cloud init error"` or similar

jenkins/scripts/utils.sh
lentzi90: Yes it works without sudo :slightly_smiling_face:

jenkins/scripts/utils.sh
lentzi90: Updated!

kashifest (approved):

furkatgofurov7 (commented):
/lgtm

PR #361, opened 2022-03-11
Add EPHEMERAL_TEST  variable for e2e test in epemeral cluster.

smoshiur1237:
EPHEMERAL_TEST variable is added to enable the CAPM3 e2e test in ephemeral cluster.  
The following jenkins pipeline [PR](https://gerrit.nordix.org/c/infra/cicd/+/12728) is created to have the CI job and use EPHEMERAL_TEST variable .


jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -13,6 +13,7 @@ export IMAGE_OS
 export DEFAULT_HOSTS_MEMORY
 export NUM_NODES
 export UPGRADE_TEST
+export EPHEMERAL_TEST
```
fmuyassarov, line 16: it is already in the vars.sh which is sourced in L8. I think this is redundant.

jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -36,6 +37,11 @@ then
   unset UPGRADE_TEST
 fi
 
+if [ "${EPHEMERAL_TEST}" == "null" ]
+then
+  unset EPHEMERAL_TEST
```
fmuyassarov, line 42: or rather set to default value `false` as in CAPM3?

smoshiur1237: I followed the same way as `UPGRADE_TEST`  and `NUM_NODES` variable.

jenkins/scripts/files/run_integration_tests.sh
smoshiur1237: I understand, but follow the same pattern like other variables such as UPGRADE_TEST

Rozzii: Yeah this section is a bit funky with the exports, but I am not sure if you should fix it here or in an other PR though.

fmuyassarov: Let's tackle this on a follow up PR @smoshiur1237 
/approve

PR #362, opened 2022-03-14
Change git diff logic in custom skip integration script

furkatgofurov7:
This PR changes the way how we find out the changes made in the PR when custom skipping integration tests are used. The problem with the current implementation is, it does not consider the case, where for example the PR branch opened some time ago falls behind the origin branch it checked out (the branch is ahead of its checked out branch X commits and behind it, Y commits), then, git diff outputs all the changes that went in with Y commits since we are doing comparison from current branch against origin/{default_branch}. 
Now, this is done in a way that we look for all changes on the current branch since the branch start point as [here](https://stackoverflow.com/questions/29810331/is-there-a-quick-way-to-git-diff-from-the-point-or-branch-origin).

### DIFF
--- a/jenkins/scripts/skip_integration_test.sh
+++ b/jenkins/scripts/skip_integration_test.sh
@@ -40,8 +40,11 @@ exclude_markdown_and_owners_files() {
 # otherwise check the updated branch to decide on whether skipping or running 
 # the integration tests.
 if [[ "${UPDATED_BRANCH}" == "${REPO_BRANCH}" ]] && [[ "${UPDATED_REPO}" == *"${REPO_ORG}/${REPO_NAME}"* ]]; then
+  echo "Main job is runnning, not skipping integration tests"
   return 1
 else
+  echo "Clone updated repo"
   gclonecd $UPDATED_REPO
+  echo "Run skipping the integration test custom script to find out git diff"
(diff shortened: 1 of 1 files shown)


lentzi90 (approved):
/lgtm
Thanks for fixing! I was just looking at this and wondering why it wanted to run the tests for one of my PRs.

furkatgofurov7:
> Thanks for fixing! I was just looking at this and wondering why it wanted to run the tests for one of my PRs.

This was caught thanks to your PR in CAPM3 :grinning:

smoshiur1237 (approved):
LGTM

namnx228 (approved):
LGTM

fmuyassarov (commented):
/hold
@furkatgofurov7 can you please squash commits before merging
/approve

furkatgofurov7:
> @furkatgofurov7 can you please squash commits before merging

I would rather prefer to keep it as is since I isolated them on purpose (1 to just change git diff logic and second for logging)

fmuyassarov:
> > @furkatgofurov7 can you please squash commits before merging
> 
> I would rather prefer to keep it as is since I isolated them on purpose (1 to just change git diff logic and second for logging)

okay
/hold cancel

PR #369, opened 2022-03-28
Temporary fix while Nordix registry is down

lentzi90:
Use upstream quay and docker hub while we wait for Nordix registry to be back up again.


kashifest:
/approve
/lgtm
Taking this in to fix the CI temporarily.

PR #370, opened 2022-03-29
add ironic-image testing pipeline

Rozzii:


jenkins/jobs/ironic_image_build_test.pipeline
```diff
@@ -0,0 +1,92 @@
...
+    OS_REGION_NAME="Kna1"
+    OS_PROJECT_NAME="Default Project 37137"
+    OS_TENANT_NAME="${env.OS_PROJECT_NAME}"
+    OS_AUTH_VERSION=3
+    OS_IDENTITY_API_VERSION="${env.OS_AUTH_VERSION}"
+    TEST_EXECUTER_VM_NAME = "${VM_NAME}"
+    TEST_EXECUTOR_VM_KEY = "${VM_KEY}"
+    # ironic-image build option(s)
+    PATCHFILE_CONTENT="${params.PATCHFILE_CONTENT ?: 'openstack/ironic refs/heads/master:refs/remotes/origin/master' }"
+    # Currnet directory
```
lentzi90, line 41:
```suggestion
    # Current directory
```

jenkins/scripts/ironic_build_test.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+
+set -eu
+
+# shellcheck disable=SC1090
+source "${VARS_SCRIPT:-/tmp/vars.sh}"
+
+# This script is used to build Ironic and it is used by multiple different pipelines.
+# This script is designed to be executed on dinamically allocated VM-s provided
```
lentzi90, line 9:
```suggestion
# This script is designed to be executed on dynamically allocated VM-s provided
```

jenkins/scripts/ironic_build_test.sh
```diff
@@ -0,0 +1,105 @@
...
+SOURCE_BRANCH="${SOURCE_BRANCH:?}"
+PATCH_FILE=patch.txt
+OS_USERNAME="${OS_USERNAME:-metal3ci}"
+PATCHFILE_CONTENT="${PATCHFILE_CONTENT:-""}"
+IRONIC_INSTALL_MODE="${IRONIC_INSTALL_MODE:-INSTALL_FROM_RPM}"
+
+# This condition will be triggered for example when the related CI tooling (project-infra PR) is being
+# tested but not a PR in the ironic-image repository.
+
+# IMPAROTANT: In case it is required to test the compatiblity between project-infra and ironic-image PRs
```
lentzi90, line 31:
```suggestion
# IMPORTANT: In case it is required to test the compatibility between project-infra and ironic-image PRs
```

jenkins/scripts/ironic_build_test.sh
```diff
@@ -0,0 +1,105 @@
...
+# then the test has to be triggered on the projec-infra PR then it should be manually stopped
+# and replayed in a way that the destination and source branches and repositories are manually modified
+# according to the inronic-image PR in question.
```
lentzi90, line 34: I'm not sure what this means. Is it for testing changes to project-infra and ironic-image at the same time? Maybe an example would help?

lentzi90 (commented):
Added a few comments below.
I saw that this runs each building method one after the other (source, rpm and patch). Would there be a point in doing it in parallel? I mean create one VM of each and then run them at the same time? (Or parameterize it so it is actually three different jobs.)

jenkins/jobs/ironic_image_build_test.pipeline
Rozzii: Done

jenkins/scripts/ironic_build_test.sh
Rozzii: Done

jenkins/scripts/ironic_build_test.sh
Rozzii: Done

Rozzii:
> Added a few comments below. I saw that this runs each building method one after the other (source, rpm and patch). Would there be a point in doing it in parallel? I mean create one VM of each and then run them at the same time? (Or parameterize it so it is actually three different jobs.)

I agree that parallel image building is desired, but I don't think that the optimal way of doing it would be to run parallel builds in multiple VMs. First of all these 3 different  builds take about 10 minutes to finish so I think it  is a bit of an overkill to use 3 VMs to spare around 5 minutes.  The other thing is that certain layers can be reused between the 3 different build so it might not take 10 minutes, we will see when the job goes live.

In my opinion it would be better if we would transition gradually to use docker compose to build  the images because that supports parallel/staged image builds and would provide more functionality anyways but 
that requires changes in the image repos and I think this is outside the scope of this PR  and my current goals. 

But I support  introducing  the docker compose in the future in the image repos.

Rozzii:
> Why can't we reuse existing integration test Jenkins pipeline + shell scripts + just introduce INSTALL_TYPE variable ? For example, this might work, but maybe I am missing something
> 
>     1. Add a new JJB with INSTALL_TYPE variable set to `source`, and add a unique trigger phrase
> 
>     2. Check INSTALL_TYPE here in integration test shell script, if source, we pass it to dev-env
>        where we already have a [logic](https://github.com/metal3-io/metal3-dev-env/blob/c5361022e334e09cf911fa53133d36bbec46700b/02_configure_host.sh#L183) to handle it, if null we fall back to rpm build.
> 
>     3. Add JJB id to ironic-image repo in projects.yaml (Gerrit)

1. The reason to have a separate code base here is that  I this test job is created for the community and tests the community images and the community will maintain it so the pipeline scripts will live here. 

2. Because the pipeline scripts live here I have  decided to test how would this work if the pipeline script is here but the actual build script is in Nordix repository and I find it less maintainable and although this build script share some commanality (very little) with the ironic build script used in the metal3-dev-tools repo combining the two in a usable way would require a lot of unnecessary code (I have actually implemented it and I didn't like it).  It is much more maintainable to have the script here then to mix it with the metal3-dev-env ironic builds script.

3. Having the build script here is not duplication this is a completely different use-case, with an almost totally different code.
 This test script is not only tests the "source" INSTALL_TYPE it tests all 3 ways of installing ironic, this script does not push the images unlike the metal3-dev-env ironic build scrip, this script does not save the metadata as the other,  this script adds parametes the patching in a different way.

jenkins/scripts/ironic_build_test.sh
Rozzii: done

fmuyassarov:
> 1. The reason to have a separate code base here is that  this test job is created for the community and tests the community images and the community will maintain it so the pipeline scripts will live here.

I didn't mean to use nordix scripts, but the pipeline that already exists in project-infra: https://github.com/metal3-io/project-infra/blob/main/jenkins/jobs/integration_tests.pipeline

> 2. Because the pipeline scripts live here I have  decided to test how would this work if the pipeline script is here but the actual build script is in Nordix repository and I find it less maintainable and although this build script share some commonality (very little) with the ironic build script used in the metal3-dev-tools repo, combining the two in a usable way would require a lot of unnecessary code (I have actually implemented it and I didn't like it).  It is much more maintainable to have the script here then to mix it with the metal3-dev-env ironic builds script.

I think we misunderstood each other. My thinking was that, If I am not wrong we would like to build Ironic image from source(sometimes) instead of RPM like we do now. For that, the only variable we need to pass is `INSTALL_TYPE`. Perhaps I understood it wrong, but I thought we want to build Ironic image from source, then utilize it within dev-env tests during a patch test that is submitted to ironic-image repo only. And for that the flow could be like
1. Trigger the job with its unique trigger phrase by leaving a comment on a PR in ironic-image repository
2. JJB will pass `INSTALL_TYPE` variable with true value
3. Jenkins pipeline in project-infra inherits the value and passed to dev-env script
4. while building Ironic image locally, evaluate `INSTALL_TYPE` and built it from source since the value is set true

Rozzii:
> > 1. The reason to have a separate code base here is that  this test job is created for the community and tests the community images and the community will maintain it so the pipeline scripts will live here.
> 
> I didn't mean to use nordix scripts, but the pipeline that already exists in project-infra: https://github.com/metal3-io/project-infra/blob/main/jenkins/jobs/integration_tests.pipeline
> 
> > 2. Because the pipeline scripts live here I have  decided to test how would this work if the pipeline script is here but the actual build script is in Nordix repository and I find it less maintainable and although this build script share some commonality (very little) with the ironic build script used in the metal3-dev-tools repo, combining the two in a usable way would require a lot of unnecessary code (I have actually implemented it and I didn't like it).  It is much more maintainable to have the script here then to mix it with the metal3-dev-env ironic builds script.
> 
> I think we misunderstood each other. My thinking was that, If I am not wrong we would like to build Ironic image from source(sometimes) instead of RPM like we do now. For that, the only variable we need to pass is `INSTALL_TYPE`. Perhaps I understood it wrong, but I thought we want to build Ironic image from source, then utilize it within dev-env tests during a patch test that is submitted to ironic-image repo only. And for that the flow could be like
> 
>     1. Trigger the job with its unique trigger phrase by leaving a comment on a PR in ironic-image repository
> 
>     2. JJB will pass `INSTALL_TYPE` variable with true value
> 
>     3. Jenkins pipeline in project-infra inherits the value and passed to dev-env script
> 
>     4. while building Ironic image locally, evaluate `INSTALL_TYPE` and built it from source since the value is set true

- I think that it is better to provide a test job that actually tests that all install types offered by the ironic-image repo will work after a change.

- I think a clean general purpose ironic-image testing pipeline is more maintainable and more useful for the community going forward.

jenkins/scripts/start_ironic_image_build_test.sh
```diff
@@ -0,0 +1,119 @@
...
+  -o UserKnownHostsFile=/dev/null \
+  -o ServerAliveInterval=15 \
+  -o ServerAliveCountMax=10 \
+  -i "${METAL3_CI_USER_KEY}" \
+  "${METAL3_CI_USER}"@"${TEST_EXECUTER_IP}" \
+  PATH="/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin" \
+  IRONIC_INSTALL_MODE="INSTALL_FROM_SOURCE" \
+  "/tmp/run_build_ironic.sh"
+
+echo "Running Ironic image building scripti (INSTALL_FROM_RPM)"
```
fmuyassarov, line 93:
```suggestion
echo "Running Ironic image building script (INSTALL_FROM_RPM)"
```

jenkins/scripts/start_ironic_image_build_test.sh
```diff
@@ -0,0 +1,119 @@
...
+ssh \
+  -o StrictHostKeyChecking=no \
+  -o UserKnownHostsFile=/dev/null \
+  -o ServerAliveInterval=15 \
+  -o ServerAliveCountMax=10 \
+  -i "${METAL3_CI_USER_KEY}" \
+  "${METAL3_CI_USER}"@"${TEST_EXECUTER_IP}" \
+  PATH="/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin" \
+  IRONIC_INSTALL_MODE="INSTALL_FROM_RPM" \
+  "/tmp/run_build_ironic.sh"
```
fmuyassarov, line 105:
```suggestion
  "/tmp/ironic_build_test.sh"
```

jenkins/scripts/start_ironic_image_build_test.sh
```diff
@@ -0,0 +1,119 @@
...
+ssh \
+  -o StrictHostKeyChecking=no \
+  -o UserKnownHostsFile=/dev/null \
+  -o ServerAliveInterval=15 \
+  -o ServerAliveCountMax=10 \
+  -i "${METAL3_CI_USER_KEY}" \
+  "${METAL3_CI_USER}"@"${TEST_EXECUTER_IP}" \
+  PATH="/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin" \
+  IRONIC_INSTALL_MODE="INSTALL_FROM_RPM_PATCH" \
+  "/tmp/run_build_ironic.sh"
```
fmuyassarov, line 119:
```suggestion
  "/tmp/ironic_build_test.sh"
```

jenkins/scripts/ironic_build_test.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# This condition will be triggered for example when the related CI tooling (project-infra PR) is being
+# tested but not a PR in the ironic-image repository.
+
+# IMPORTANT: In case it is required to test the compatibility between project-infra and ironic-image PRs
+# then the test has to be triggered on the projec-infra PR then it should be manually stopped
+# and replayed in a way that the destination and source branches and repositories are manually modified
+# according to the inronic-image PR in question.
+# Example: user have in both project-infra and in ironic-image and the uswer would like to test both
+# together before any of them would be merged.
```
fmuyassarov, line 36:
```suggestion
# Example: user have in both project-infra and in ironic-image and the user would like to test both
# together before any of them would be merged.
```

jenkins/scripts/ironic_build_test.sh
```diff
@@ -0,0 +1,107 @@
...
+if [[ "${DESTINATION_REPO}" != "https://github.com/metal3-io/ironic-image.git" ]];then
+    DESTINATION_REPO="https://github.com/metal3-io/ironic-image.git"
+    DESTINATION_BRANCH="main"
+    SOURCE_REPO="${DESTINATION_REPO}"
+    SOURCE_BRANCH="${DESTINATION_BRANCH}"
+fi
+
+case "$IRONIC_INSTALL_MODE" in
+
+  'INSTALL_FROM_SOURCE')
```
fmuyassarov, line 46: maybe don't repeat `prepare_work_env` take it out of case?

Rozzii: It is done this way to avoid work environment preparation in case unsupported IRONIC_INSTALL_MODE mode was supplied to the script.

jenkins/jobs/ironic_image_build_test.pipeline
```diff
@@ -0,0 +1,92 @@
...
+           currentBuild.result = 'FAILURE'
+         }
+       }
+    }
+    cleanup {
+      withCredentials([usernamePassword(credentialsId: 'metal3ci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')]) {
+        withCredentials([sshUserPrivateKey(credentialsId: 'metal3ci_city_cloud_ssh_keypair', keyFileVariable: 'METAL3_CI_USER_KEY')]){
+          script {
+            if ("${SKIP_DELETION}" == "true") {
+              echo "Skipping VM deletion"
```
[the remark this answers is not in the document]
Rozzii: done

jenkins/jobs/ironic_image_build_test.pipeline
```diff
@@ -0,0 +1,92 @@
+import java.text.SimpleDateFormat
+
+// 2 hours
+def TIMEOUT = 7200
```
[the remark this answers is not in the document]
Rozzii: done

jenkins/jobs/ironic_image_build_test.pipeline
```diff
@@ -0,0 +1,92 @@
...
+
+script {
+  def date = new Date()
+  def dateFormat = new SimpleDateFormat("yyyyMMddHHmmss")
+  def rand = new Random()
+  VM_KEY = (1..4).collect { ('a'..'z').join("")[ rand.nextInt( 26 ) ] }.join("")
+  VM_NAME = "ci-builder-vm-" + dateFormat.format(date) + "-" + VM_KEY
```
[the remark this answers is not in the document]
Rozzii: done

jenkins/jobs/ironic_image_build_test.pipeline
```diff
@@ -0,0 +1,92 @@
...
+    PATCHFILE_CONTENT="${params.PATCHFILE_CONTENT ?: 'openstack/ironic refs/heads/master:refs/remotes/origin/master' }"
+    # Current directory
+    CURRENT_DIR = sh (
+                      script: 'readlink -f "."',
+                      returnStdout: true
+                     ).trim()
+  }
+
+  stages {
+    stage('Build ironic-image container image'){
```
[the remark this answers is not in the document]
Rozzii: done

jenkins/scripts/start_ironic_image_build_test.sh
Rozzii: done

jenkins/scripts/start_ironic_image_build_test.sh
```diff
@@ -0,0 +1,119 @@
...
+PATCHFILE_CONTENT="${PATCHFILE_CONTENT}"
+EOF
+
+
+# Send Ironic script to remote executer
+scp \
+  -o StrictHostKeyChecking=no \
+  -o UserKnownHostsFile=/dev/null \
+  -i "${METAL3_CI_USER_KEY}" \
+  "${CI_DIR}/run_build_ironic.sh" \
```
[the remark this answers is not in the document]
Rozzii: done

jenkins/scripts/start_ironic_image_build_test.sh
```diff
@@ -0,0 +1,119 @@
...
+ssh \
+  -o StrictHostKeyChecking=no \
+  -o UserKnownHostsFile=/dev/null \
+  -o ServerAliveInterval=15 \
+  -o ServerAliveCountMax=10 \
+  -i "${METAL3_CI_USER_KEY}" \
+  "${METAL3_CI_USER}"@"${TEST_EXECUTER_IP}" \
+  PATH="/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin" \
+  IRONIC_INSTALL_MODE="INSTALL_FROM_SOURCE" \
+  "/tmp/run_build_ironic.sh"
```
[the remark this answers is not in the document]
Rozzii: done

jenkins/scripts/start_ironic_image_build_test.sh
Rozzii: done

jenkins/scripts/start_ironic_image_build_test.sh
Rozzii: done

jenkins/scripts/ironic_build_test.sh
Rozzii: done

Rozzii:
/hold
waiting for this https://github.com/metal3-io/ironic-image/pull/368

jenkins/scripts/ironic_build_test.sh
fmuyassarov: For that you could add a check to see if the user has provided unsupported method and fail early

fmuyassarov (commented):
Looks good :+1:

jenkins/scripts/ironic_build_test.sh
Rozzii: This is the earliest part in this script where any actual "logic" is implemented. If you mean that I could do it outside this script then you are right but I have implemented the script to be usable in a standalone mode in case somebody has to quickly test something on local machine.

jenkins/scripts/start_ironic_image_build_test.sh
```diff
@@ -0,0 +1,119 @@
...
+  -o UserKnownHostsFile=/dev/null \
+  -o ServerAliveInterval=15 \
+  -o ServerAliveCountMax=10 \
+  -i "${METAL3_CI_USER_KEY}" \
+  "${METAL3_CI_USER}"@"${TEST_EXECUTER_IP}" \
+  PATH="/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin" \
+  IRONIC_INSTALL_MODE="INSTALL_FROM_RPM" \
+  "/tmp/ironic_build_test.sh"
+
+echo "Running Ironic image building scripti (INSTALL_FROM_RPM_PATCH)"
```
elfosardo, line 107: nit: s/scripti/script

elfosardo (approved):
thanks!

jenkins/scripts/start_ironic_image_build_test.sh
Rozzii: done

lentzi90 (commented):
/lgtm
Nice work on this @Rozzii !

Rozzii:
Test build has been successful https://jenkins.nordix.org/view/Metal3/job/metal3_project_infra_ironic_image_build_test/22/console

lentzi90 (commented):
/lgtm

PR #373, opened 2022-04-13
enable ironic source install for integration tests

Rozzii:


fmuyassarov (commented):
If I understand it right whoever is triggering the job can't explicitly mention that image should be build from source right? What's the use of those 3 variables ?

Rozzii:
> If I understand it right whoever is triggering the job can't explicitly mention that image should be build from source right? What's the use of those 3 variables ?

I have configured the JJB in a way that the variables will be populated with "ghprb" values, so the workflow will be 
(push PR -> call e.g. "/test-v1b1-centos-integration-ironic-source" -> job gets triggered and the github plugin should use the PR's "ghprb" values to populate the variables in question.

jenkins/scripts/integration_test.sh
```diff
@@ -43,6 +43,14 @@ EPHEMERAL_TEST="${EPHEMERAL_TEST:-false}"
 TEST_EXECUTER_VM_NAME="${TEST_EXECUTER_VM_NAME}"
 TEST_EXECUTER_PORT_NAME="${TEST_EXECUTER_PORT_NAME:-${TEST_EXECUTER_VM_NAME}-int-port}"
 TEST_EXECUTER_FIP_TAG="${TEST_EXECUTER_FIP_TAG:-${TEST_EXECUTER_VM_NAME}-floating-ip}"
+IRONIC_INSTALL_TYPE="${IRONIC_INSTALL_TYPE:-rpm}"
+IRONIC_FROM_SOURCE="${IRONIC_FROM_SOURCE:-false}"
+IRONIC_LOCAL_IMAGE_BRANCH="${IRONIC_LOCAL_IMAGE_BRANCH:-}"
+IRONIC_LOCAL_IMAGE="${IRONIC_LOCAL_IMAGE:-}"
```
lentzi90, line 49: Why do we need these? If the variable is set, it already has the value, otherwise we set it to empty...?

fmuyassarov (commented):
/approve
But I agree with @lentzi90's comment.

fmuyassarov:
Ow I missed that there is lgtm already :)

PR #375, opened 2022-04-21
Remove workaround for "no-space" issue

lentzi90:
The issue should now be fixed and we can rely on checking cloud-init status instead of comparing filesystem and disk sizes. The retry for server creation was also removed since we do not expect there to be issues that can be solved by recreating the server anymore.

### DIFF
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -108,39 +108,13 @@ then
   "$FLOATING_IP"
 fi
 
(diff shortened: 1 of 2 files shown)


lentzi90:
/hold
Testing https://github.com/Nordix/metal3-dev-tools/pull/516 by using hard coded image.

kashifest (approved):

fmuyassarov (commented):
/lgtm

PR #376, opened 2022-04-22
fix ironic source install in integration tests

Rozzii:
The previous implementation relied on jenkins parameter templates that
in fact can't be correctly substituted with "ghprb" data because of the
limitations of jenkins build startup process.


lentzi90 (commented):
/lgtm

lentzi90 (commented):
/lgtm

jenkins/scripts/integration_test.sh
```diff
@@ -45,11 +45,20 @@ TEST_EXECUTER_PORT_NAME="${TEST_EXECUTER_PORT_NAME:-${TEST_EXECUTER_VM_NAME}-int
...
+if [ "${IRONIC_INSTALL_TYPE}" == "source" ];
+then
     IRONIC_FROM_SOURCE="true"
+    if [ "${REPO_NAME}" == "ironic-image" ];
+    then
+        IRONIC_LOCAL_IMAGE="${ghprbAuthorRepoGitUrl}"
+        IRONIC_LOCAL_IMAGE_BRANCH="${ghprbActualCommit}"
+    else
+        IRONIC_LOCAL_IMAGE="https://github.com/metal3-io/ironic-image.git"
+        IRONIC_LOCAL_IMAGE_BRANCH="main"
```
furkatgofurov7, line 60: this is when `install type = source` and not `repo != ironic-image` repo, and should not it be outside of outer condition?

Rozzii: outside the outer condition is `install type != source` and in that case  both variable will be empty.
Inside the first condition it could be `install type = source` and` repo != ironic-image` so I can spare the `else` branch and move the value overrides under directly `IRONIC_FROM_SOURCE="true"` but at that case we would trade an `else` block for a double override and I am not sure it worth a fix or is it even matters.

furkatgofurov7: okay

furkatgofurov7 (approved):
/approve

PR #378, opened 2022-04-29
Add side by side e2e feature tests

mboukhalfa:
Create new pipeline to run e2e features test in parallel VMs and add the necessary changes in the scripts to avoid conflicts when preparing the vars.sh file for each VM

### DIFF
--- a/jenkins/jobs/parallel_e2e_features_test.pipeline
+++ b/jenkins/jobs/parallel_e2e_features_test.pipeline
@@ -1,0 +1,240 @@
+import java.text.SimpleDateFormat
+
+ci_git_credential_id = "metal3-jenkins-github-token"
+
+// 10 minutes
+def CLEAN_TIMEOUT = 600
+// 3 hours
+def TIMEOUT = 10800
+SKIP_DELETION = (env.SKIP_DELETION)
+
+script {
+  if ("${PROJECT_REPO_ORG}" == "metal3-io" && "${PROJECT_REPO_NAME}" == "project-infra") {
+    echo "Checkout ${ghprbAuthorRepoGitUrl} branch ${ghprbActualCommit}"
+    ci_git_branch = "${ghprbActualCommit}"
+    ci_git_url = "${ghprbAuthorRepoGitUrl}"
+  } else {
+    echo "Checkout ${ghprbAuthorRepoGitUrl} main"
+    ci_git_branch = "main"
+    ci_git_url = "https://github.com/metal3-io/project-infra.git"
+  }
+
+  def date = new Date()
+  def dateFormat = new SimpleDateFormat("yyyyMMddHHmmss")
+  def rand = new Random()
+  VM_KEY = (1..4).collect {
+    ('a'..
+      'z').join("")[rand.nextInt(26)]
+  }.join("")
+
+  if ("${SKIP_DELETION}" == "true") {
+    VM_NAME = "ci-test-keep-vm-" + dateFormat.format(date) + "-" + VM_KEY
+  } else {
+    VM_NAME = "ci-test-vm-" + dateFormat.format(date) + "-" + VM_KEY
+  }
+}
+
+pipeline {
+  agent {
+    label 'metal3-static-workers'
+  }
+  environment {
+    METAL3_CI_USER = "metal3ci"
+    REPO_ORG = "${PROJECT_REPO_ORG}"
+    REPO_NAME = "${PROJECT_REPO_NAME}"
+    UPDATED_REPO = "${ghprbAuthorRepoGitUrl}"
+    REPO_BRANCH = "${ghprbTargetBranch}"
+    UPDATED_BRANCH = "${ghprbActualCommit}"
+    OS_USERNAME = "metal3ci"
+    OS_AUTH_URL = "https://kna1.citycloud.com:5000"
+    OS_USER_DOMAIN_NAME = "CCP_Domain_37137"
+    OS_PROJECT_DOMAIN_NAME = "CCP_Domain_37137"
+    OS_REGION_NAME = "Kna1"
+    OS_PROJECT_NAME = "Default Project 37137"
+    OS_TENANT_NAME = "Default Project 37137"
+    OS_AUTH_VERSION = 3
+    OS_IDENTITY_API_VERSION = 3
+    TEST_EXECUTER_VM_NAME = "${VM_NAME}"
+    BUILD_TAG = "${env.BUILD_TAG}"
+    PR_ID = "${ghprbPullId}"
+    IMAGE_OS = "${IMAGE_OS}"
+    CAPI_VERSION = "${CAPI_VERSION}"
+    CAPIRELEASE = "${CAPIRELEASE}"
+    CAPM3_VERSION = "${CAPM3_VERSION}"
+    CAPM3RELEASE = "${CAPM3RELEASE}"
+    DEFAULT_HOSTS_MEMORY = "${TARGET_NODE_MEMORY}"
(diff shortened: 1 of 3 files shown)


mboukhalfa:
pipe broken fail 
/test-v1b1-centos-e2e-parallel

mboukhalfa:
pipe broken failure again
/test-v1b1-centos-e2e-parallel

jenkins/scripts/integration_test.sh
```diff
@@ -124,8 +125,13 @@ if ! vm_healthy "${METAL3_CI_USER}" "${METAL3_CI_USER_KEY}" "${TEST_EXECUTER_IP}
...
 fi
-
-cat <<-EOF >> "${CI_DIR}/files/vars.sh"
+# We should not create vars.sh on the static worker because it will be overridden when running parallel 
+# tests since GINKGO_FOCUS value is different for each parallel stage
+cat <<-EOF | ssh \
+  -o StrictHostKeyChecking=no \
+  -o UserKnownHostsFile=/dev/null \
+  -i "${METAL3_CI_USER_KEY}" \
+  "${METAL3_CI_USER}@${TEST_EXECUTER_IP}" -T "cat > /tmp/vars.sh"
```
kashifest, line 134: This part and the comment above is not clear to me.

jenkins/scripts/integration_test.sh
```diff
@@ -159,13 +167,19 @@ fi
 
 cat "${CI_DIR}/integration_test_env.sh" >> "${CI_DIR}/files/vars.sh"
 
+# Append to vars.sh in the remote vm
```
lentzi90, line 170: Why not write the whole file here first and then copy to the remote VM? Now we copy it over in two parts and have only half of it on the jenkins worker. This seems strange to me.

jenkins/jobs/parallel_e2e_features_test.pipeline
```diff
@@ -0,0 +1,227 @@
...
+
+              withCredentials([usernamePassword(credentialsId: 'metal3ci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')]) {
+                withCredentials([sshUserPrivateKey(credentialsId: 'metal3ci_city_cloud_ssh_keypair', keyFileVariable: 'METAL3_CI_USER_KEY')]) {
+                  withCredentials([string(credentialsId: 'metal3-clusterctl-github-token', variable: 'GITHUB_TOKEN')]) { 
+                    ansiColor('xterm') {
+                      sh "./jenkins/scripts/integration_test.sh"
+                    }
+                  }
+                }
+              }
```
lentzi90, line 111: I think it would be easier to read if we skip the nested `withCredentials` and just list all credentials in one. Same for the other steps. (We have them nested in all the old pipelines but why not do it like this for new ones. :slightly_smiling_face: )
```suggestion
              withCredentials([usernamePassword(credentialsId: 'metal3ci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD'),
                sshUserPrivateKey(credentialsId: 'metal3ci_city_cloud_ssh_keypair', keyFileVariable: 'METAL3_CI_USER_KEY'),
                string(credentialsId: 'metal3-clusterctl-github-token', variable: 'GITHUB_TOKEN')]) {
                    ansiColor('xterm') {
                      sh "./jenkins/scripts/integration_test.sh"
                    }
              }
```

jenkins/scripts/integration_test.sh
mboukhalfa: If we write the file on the disk it will be overridden by the parallel test  which means will be moving wrong file content to one of the VMs, the part written on the disk is not conflicting but the other we write it directly on the target VM

lentzi90: Ok, so this part is the same for both VMs? I think it would be better then to write to a random temp file. We could still get race conditions here even if the content is the same. For example, we could try to copy the file while the parallel job is still writing to it so we get half a file.

I would suggest to use `mktemp` to create a file instead of using `vars.sh`. Then we avoid the parallelism issues and can still write everything first before we copy it over. Something like `mktemp vars-XXXXXX.sh` should do the trick.

jenkins/jobs/parallel_e2e_features_test.pipeline
mboukhalfa: looks better now thanks

jenkins/scripts/integration_test.sh
mboukhalfa: done thanks

jenkins/jobs/parallel_e2e_features_test.pipeline
```diff
@@ -0,0 +1,239 @@
...
+          steps {
+            withCredentials([usernamePassword(credentialsId: 'metal3ci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')] {
+              [sshUserPrivateKey(credentialsId: 'metal3ci_city_cloud_ssh_keypair', keyFileVariable: 'METAL3_CI_USER_KEY')] {
+                [string(credentialsId: 'metal3-clusterctl-github-token', variable: 'GITHUB_TOKEN')] {
+                  ansiColor('xterm') {
+                    sh "./jenkins/scripts/integration_test.sh"
+                  }
+                }
+              }
+            })
```
lentzi90, line 123: Sorry I was probably unclear before. I mean we don't need so many brackets. Something like this:

```
  withCredentials([
     usernamePassword(credentialsId: 'metal3ci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD'),
     sshUserPrivateKey(credentialsId: 'metal3ci_city_cloud_ssh_keypair', keyFileVariable: 'METAL3_CI_USER_KEY'),
     string(credentialsId: 'metal3-clusterctl-github-token', variable: 'GITHUB_TOKEN')
    ] {
      ansiColor('xterm') {
        sh "./jenkins/scripts/integration_test.sh"
      }
  })
```

So only one `[]` where all credentials goes like `[ credential1, credential2]`.

jenkins/scripts/integration_test.sh
```diff
@@ -143,30 +145,36 @@ EPHEMERAL_TEST="${EPHEMERAL_TEST}"
...
-  "${CI_DIR}/files/vars.sh" \
-  "${METAL3_CI_USER}@${TEST_EXECUTER_IP}:/tmp/" > /dev/null
+  "${METAL3_CI_USER}@${TEST_EXECUTER_IP}:/tmp" > /dev/null
+
+scp \
+  -o StrictHostKeyChecking=no \
+  -o UserKnownHostsFile=/dev/null \
+  -i "${METAL3_CI_USER_KEY}" \
+  "${CI_DIR}/files/${TEMP_FILE_NAME}" \
+  "${METAL3_CI_USER}@${TEST_EXECUTER_IP}:/tmp/vars.sh" > /dev/null
```
Rozzii, line 178: Here after you have copied everything, I would suggest to delete the temp var file from the static worker.

lentzi90 (commented):
/lgtm

Rozzii (approved):
Looks good to me, just a minor optional suggestion.

jenkins/scripts/integration_test.sh
mboukhalfa: done

lentzi90 (commented):
/lgtm

kashifest:
/assign @fmuyassarov 
/approve
/hold 
Until @fmuyassarov reviews it

PR #379, opened 2022-05-05
WIP increase ssh timeout in tests

Rozzii:
This modification should increase ssh timeout when running,
integration and feature tests (e2e).

### DIFF
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -174,8 +174,7 @@ echo "Running the tests"
 ssh \
   -o StrictHostKeyChecking=no \
   -o UserKnownHostsFile=/dev/null \
-  -o ServerAliveInterval=15 \
-  -o ServerAliveCountMax=10 \
+  -o ServerAliveInterval=10800 \
   -i "${METAL3_CI_USER_KEY}" \
   "${METAL3_CI_USER}"@"${TEST_EXECUTER_IP}" \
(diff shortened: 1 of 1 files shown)


Rozzii:
//test-v1b1-centos-e2e

jenkins/scripts/integration_test.sh
```diff
@@ -174,8 +174,7 @@ echo "Running the tests"
 ssh \
   -o StrictHostKeyChecking=no \
   -o UserKnownHostsFile=/dev/null \
-  -o ServerAliveInterval=15 \
-  -o ServerAliveCountMax=10 \
+  -o ServerAliveInterval=10800 \
```
mboukhalfa, line 177: This is 3h ? does it mean that the server will try to reconnect for 3h after broken pipe

Rozzii: No, this means that the server waits for 3 hours after the last char was sent on ssh.

Rozzii: Broken pipe can happen simply because the connection times out because there is no communication on the connection. 
This situation can easily happen if the test takes a lot of time and it does not print anything to the console.

mboukhalfa:
We close since the issue was fixed in https://github.com/Nordix/metal3-dev-tools/pull/556

Rozzii:
Issue was solved differently.
/close

PR #381, opened 2022-05-05
Exporting M3_DEV_ENV_PATH to e2e

mboukhalfa:
When testing from` dev-env`, e2e should find it cloned in `tested_repo` instead of cloning it under `/opt/metal3-dev-env`
e2e clone dev-env here https://github.com/metal3-io/cluster-api-provider-metal3/blob/62562ec2522e3c4afa541cc75b1e9c111263a120/scripts/ci-e2e.sh#L19

### DIFF
--- a/jenkins/scripts/integration_test_env.sh
+++ b/jenkins/scripts/integration_test_env.sh
@@ -6,6 +6,7 @@ if [ "${REPO_NAME}" == "metal3-dev-env" ]
 then
   export METAL3REPO="${UPDATED_REPO}"
   export METAL3BRANCH="${UPDATED_BRANCH}"
+  export M3_DEV_ENV_PATH="${HOME}/tested_repo"
 
   # If the target repo and branch are the same as the source repo and branch
   # we're running a main test, that is not for a PR, so we build the image


lentzi90 (commented):
This works only when `FORCE_REPO_UPDATE` is `false`, so I wonder if we should also explicitly set it here? What do you think?
Currently it is false by default for the e2e tests in CAPM3, but if we set it here also it may be a bit more obvious what's going on.

mboukhalfa:
> This works only when `FORCE_REPO_UPDATE` is `false`, so I wonder if we should also explicitly set it here? What do you think? Currently it is false by default for the e2e tests in CAPM3, but if we set it here also it may be a bit more obvious what's going on.

I see it explicitly set in line 47 so I do not thing we need to repeat that inside the if

lentzi90 (commented):
> I see it explicitly set in line 47 so I do not think we need to repeat that inside the if

Nice, didn't see that!
\lgtm

mboukhalfa:
> > I see it explicitly set in line 47 so I do not think we need to repeat that inside the if
> 
> Nice, didn't see that! \lgtm

Now I think that changes need to be in integration_test_env.sh where we are exporting all the other paths like BMOPATH ..

lentzi90 (commented):
/lgtm

PR #383, opened 2022-05-11
Add deletion of orphaned Fra1 floating IPs

Rozzii:
This commit adds openstack commands to check and if possible
delete orphaned floating IPs from Fra1 region.

### DIFF
--- a/jenkins/scripts/integration_delete.sh
+++ b/jenkins/scripts/integration_delete.sh
@@ -38,6 +38,9 @@ then
   echo "Deleting executer floating IP ${TEST_EXECUTER_FIP_ID}."
   echo "${TEST_EXECUTER_FIP_ID}" | xargs openstack floating ip delete
   echo "Executer floating IP ${TEST_EXECUTER_FIP_ID} is deleted."
+
+  # Check and delete orphaned floating IPs
+  openstack floating ip list --status "DOWN" --column "ID" -f json | jq --raw-output '.[]."ID"' | xargs -0 openstack floating ip delete
 fi
 
 # Delete executer vm


jenkins/scripts/integration_delete.sh
```diff
@@ -40,6 +40,9 @@ then
   echo "Executer floating IP ${TEST_EXECUTER_FIP_ID} is deleted."
 fi
 
+# Check and delete orphaned floating IPs
+openstack floating ip list --status "DOWN" --column "ID" -f json | jq --raw-output '.[]."ID"'| xargs -0 openstack floating ip delete
```
lentzi90, line 44: nit
```suggestion
openstack floating ip list --status "DOWN" --column "ID" -f json | jq --raw-output '.[]."ID"' | xargs -0 openstack floating ip delete
```

lentzi90 (commented):
One nit, otherwise this is good!

Rozzii:
> One nit, otherwise this is good!
> 
> Edit: Should we run this in both Kna1 and Fra1 or just Fra1?

Only for Fra1, we don't use floating IPs for the test jobs in Kna1. Floating IPs are created automatically in Fra1 because otherwise the static workers that live in Kna1 would not be able to connect to Fra1 ephemeral workers.

lentzi90:
> Only for Fra1, we don't use floating IPs for the test jobs in Kna1. Floating IPs are created automatically in Fra1 because otherwise the static workers that live in Kna1 would not be able to connect to Fra1 ephemeral workers.

Should we not move it into the if statement above then? Now I think it will run on both Fra1 and Kna1.

Rozzii:
> > Only for Fra1, we don't use floating IPs for the test jobs in Kna1. Floating IPs are created automatically in Fra1 because otherwise the static workers that live in Kna1 would not be able to connect to Fra1 ephemeral workers.
> 
> Should we not move it into the if statement above then? Now I think it will run on both Fra1 and Kna1.

Sorry you are right it was a mistake I wanted to put it into the condition, good catch thanks.

lentzi90 (commented):
/lgtm

smoshiur1237 (approved):
lgtm

PR #387, opened 2022-05-18
Fix e2e to not focus on empty string

mboukhalfa:
Fix skipping all e2e tests  when` ginkgo.focus="''"`


mboukhalfa:
Pipe broken ! 
/test-v1b1-centos-e2e

kashifest:
Did the tests passed?

mboukhalfa:
> Did the tests passed?

passed the stage where it was skipping all the specs but failed later with pipe broken

PR #389, opened 2022-05-20
Fix fetching e2e logs

mboukhalfa:
The condition `'[' cluster-api-provider-metal3 == 'cluster-api-provider-*' ']'`  [here](https://github.com/metal3-io/project-infra/blob/5ae9c8bcaccfb092350301c35dd7f446478c96de/jenkins/scripts/files/run_integration_tests.sh#L87)  was resulting in `false` while it should be `true`  and this because the pattern matching works only with double brackets so we need to use double `[[` instead of` [ `

Also this PR add the correct path to fetch logs when the e2e triggered from other repos


furkatgofurov7:
Let's run centos jobs from now until Ubuntu related changes merged in CAPM3. 

/test-v1b1-centos-e2e

furkatgofurov7 (commented):
/lgtm

PR #391, opened 2022-05-23
Set TCPKeepAlive to no on ssh client and server side

Sunnatillo:
Increase ssh timeout on both client and server side


kashifest (approved):

smoshiur1237 (commented):
/lgtm

PR #395, opened 2022-05-27
Handle floating ip deletion error

Rozzii:
This commit makes the Fra1 floating IP deletion command a non
blocking action. This change was needed because there could be
provider side issues that might block deltion or there could be no
IP to delete and that would also cause an error response that would
break the VM deletion workflow.


furkatgofurov7 (approved):
Thanks!

/approve

lentzi90 (approved):

PR #402, opened 2022-06-15
Fix running integration tests from capm3

mboukhalfa:
When running integration tests from capm3 the capm3 folder is cloned instead of dev-env

### DIFF
--- a/jenkins/scripts/files/run_integration_tests.sh
+++ b/jenkins/scripts/files/run_integration_tests.sh
@@ -79,11 +79,14 @@ if [[ ("${TESTS_FOR}" != "e2e_tests" && "${REPO_NAME}" == "metal3-dev-env") ||
   # If we are testing ansible test from metal3-dev-env or e2e from capm3,
   # it will already be cloned to tested_repo
   pushd tested_repo
-elif [[ "${REPO_NAME}" == "cluster-api-provider-metal3" ]]; then
+elif [[ "${TESTS_FOR}" == "e2e_tests" ]]; then
+  # only if the test is e2e clone capm3 and run the test from there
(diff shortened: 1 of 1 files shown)


jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -79,7 +79,7 @@ if [[ ("${TESTS_FOR}" != "e2e_tests" && "${REPO_NAME}" == "metal3-dev-env") ||
   # If we are testing ansible test from metal3-dev-env or e2e from capm3,
   # it will already be cloned to tested_repo
   pushd tested_repo
-elif [[ "${REPO_NAME}" == "cluster-api-provider-metal3" ]]; then
+elif [[ "${TESTS_FOR}" == "e2e_tests" ]]; then
```
lentzi90, line 82: Could you add a comment explaining this case? I guess it is when we are running e2e but not from CAPM3, right?

mboukhalfa: Currently when we run integration tests from capm3 it clones capm3 repo instead of dev-env check https://jenkins.nordix.org/job/metal3_capm3_v1b1_integration_test_ubuntu/275/

mboukhalfa: done

lentzi90 (commented):
/lgtm

mboukhalfa:
> /test-v1b1-ubuntu-integration

The issue is that we could not test triggering tests from other repos here from project infra so I am pretty sure testing here does not ensure that it will work from other repos and for this reason I missed that case in the previous PR

furkatgofurov7 (approved):
/approve

PR #405, opened 2022-06-27
Update the name of BMHs in CRs

kashifest:
changing eselda13u31s03.est.lan to eselda13u31s03. Recently BML main jobs are failing due to regex complaining about the provider ID format which containe BMH name. This PR attempts to simplify BMH name to avoid having special characters like `.` in the name. '

### DIFF
--- a/jenkins/scripts/bare_metal_lab/templates/bmhosts_crs.yaml.j2
+++ b/jenkins/scripts/bare_metal_lab/templates/bmhosts_crs.yaml.j2
@@ -20,7 +20,7 @@ data:
 apiVersion: metal3.io/v1alpha1
 kind: BareMetalHost
 metadata:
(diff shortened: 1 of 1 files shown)


lentzi90:
Why is this needed? Is it not ok to have `.` in the name?

kashifest:
> Why is this needed? Is it not ok to have `.` in the name?

Thats the only change we have, its worth testing it IMO

kashifest:
> LGTM if this fixes the issue on the ci will be good to have some context in the PR description

Description added

lentzi90 (commented):
Worth a try. Let's hope we don't get failure in the main job again
/lgtm

smoshiur1237 (approved):
/lgtm

furkatgofurov7 (approved):

PR #407, opened 2022-06-29
Fetch init container logs

### LINKED ISSUE #328: Collect logs from initContainers
Currently we only collect logs from the "normal" containers in each Pod (i.e. the ones defined in `spec.containers`). It would be good to also get the logs for init containers since it can sometimes happen that they get stuck.

Make the [run_fetch_logs.sh](https://github.com/metal3-io/project-infra/blob/main/jenkins/scripts/files/run_fetch_logs.sh) script collect logs from `spec.initContainers` also.

lentzi90:
Fixes #328


lentzi90:
You can see the init container logs from IPA downloader if you check the management cluster logs.

kashifest (approved):

furkatgofurov7 (commented):
/lgtm

PR #410, opened 2022-06-30
Fix ginkgo_focus null string value

mboukhalfa:
the script writes `'null'` in the vars file when ginkgo_focus is undefined


smoshiur1237 (approved):
/lgtm

lentzi90 (commented):
/lgtm

PR #425, opened 2022-08-01
Fix incorrect metal3 dev env path

Rozzii:
The cleanup script worked with the wron metal3 dev env path.

### DIFF
--- a/jenkins/scripts/files/run_clean.sh
+++ b/jenkins/scripts/files/run_clean.sh
@@ -6,6 +6,9 @@ REPO_NAME="${1:-metal3-dev-env}"
 IMAGE_OS="${2:-ubuntu}"
 TESTS_FOR="${3:-e2e_tests}"
 
+# shellcheck disable=SC1090
+source "/tmp/vars.sh"
+
 if [ "${IMAGE_OS}" == "ubuntu" ]; then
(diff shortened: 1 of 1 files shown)


jenkins/scripts/files/run_clean.sh
```diff
@@ -6,14 +6,18 @@ REPO_NAME="${1:-metal3-dev-env}"
 IMAGE_OS="${2:-ubuntu}"
 TESTS_FOR="${3:-e2e_tests}"
 
+if [ "${REPO_NAME}" == "metal3-dev-env" ]; then
```
mboukhalfa, line 9: I guess we do not need the if condition we can directly source the vars.sh file

Rozzii: Well we can I was just thinking that in case of other repos it is not necessary . Although you are right in theory every repo that uses this shell script should have a "/tmp/vars.sh" so I guess I can remove the condition.

Sunnatillo (commented):
LGTM

adilGhaffarDev:
looks good to me

smoshiur1237:
LGTM 
/assign @fmuyassarov

fmuyassarov (approved):

PR #429, opened 2022-08-18
Variables for BML hosts IP, ID and MAC

lentzi90:
This changes the old node_ips list into bare_metal_hosts, that is a list
of maps containing IDs, IPs and MACs of the BMHs. This way we can use
the list to render the BMH custom resources and avoid hard coding
details in the template.

The goal is to make it easier to switch between what hosts are used.
This can now be used by commenting/uncommenting items in the list,
rather than change the template and variables.


Sunnatillo (approved):
/lgtm
Good work

PR #434, opened 2022-09-14
Add cleanup script for bml lab to run after successful integration test

Sunnatillo:
This PR adds a cleanup script which cleans baremetal lab after successful integration test run and powers off bmhs


Sunnatillo:
@lentzi90

jenkins/scripts/bare_metal_lab/cleanup-lab.yaml
```diff
@@ -0,0 +1,12 @@
...
+- hosts: localhost
+  environment:
+    EPHEMERAL_CLUSTER: "{{ EPHEMERAL_CLUSTER }}"
+    EXTERNAL_VLAN_ID: "{{ EXTERNAL_VLAN_ID }}"
+  vars_files:
+    - default_vars/vars.yaml
+  tasks:
+  - name: Cleanup Bare Metal Lab
+    vars:
+      action_deploy: false
```
lentzi90, line 11: I'm not understanding this variable. What is it for?
When it is set to false, we skip cleaning logs and folders. Also skip cloning the repo?
Maybe it would be better to separate the tasks into different files instead and just include them that we want, instead of disabling tasks based on this variable.

lentzi90 (commented):
A bit confusing but I'm not sure how to do it better.
/lgtm

jenkins/scripts/bare_metal_lab/cleanup-lab.yaml
Sunnatillo: removed the variable and sorting tasks with tags now

mboukhalfa:
> cleans baremetal lab after successful integration test run and powers off bmhs

From skimming that I see that were some refactoring also can we wrap up all the changes done and update the description

Sunnatillo:
> > cleans baremetal lab after successful integration test run and powers off bmhs
> 
> From skimming that I see that were some refactoring also can we wrap up all the changes done and update the description

done.

smoshiur1237 (approved):
LGTM

jenkins/scripts/bare_metal_lab/cleanup-lab.yaml
```diff
@@ -0,0 +1,10 @@
+---
+- hosts: localhost
+  environment:
+    EPHEMERAL_CLUSTER: "{{ EPHEMERAL_CLUSTER }}"
+    EXTERNAL_VLAN_ID: "{{ EXTERNAL_VLAN_ID }}"
+  vars_files:
+    - default_vars/vars.yaml
```
furkatgofurov7, line 7: why do we need to set vars for cleanup, are not they already set in deploy-lab.yaml?

jenkins/scripts/bare_metal_lab/tasks/cleanup-tasks.yaml
```diff
@@ -0,0 +1,128 @@
+- name: Check required env vars are set
```
furkatgofurov7, line 1: I am not sure if I do agree with the name of this file vs what it does. I think we could still move back some parts of it to original deploy-lab.yaml and leave it there and leave here only the parts that are really doing the "cleaning up" tasks.

jenkins/scripts/bare_metal_lab/tasks/cleanup-tasks.yaml
```diff
@@ -0,0 +1,128 @@
...
+    - /home/{{ ansible_user_id }}/go
+    - /home/{{ ansible_user_id }}/tested_repo
+    - /home/{{ ansible_user_id }}/metal3
+    - /home/{{ ansible_user_id }}/metal3-dev-env
+    - /opt/metal3-dev-env
+    - /tmp/mariadb.csr
+    - /tmp/ironic.csr
+    - /tmp/manifests
+    - "{{ serial_log_location }}"
+  tags: deploy
```
furkatgofurov7, line 72: if I understood right `--skip-tags "deploy"` in bml_cleanup.sh means we would skip the tasks with this tag, however we do need these to be run as part of the cleanup, are not we? also, `tags:deploy` inside the cleanup_tasks.yaml looks bit strange to me, how about renaming it to `tags:cleanup`?

Sunnatillo: There are some cleanup tasks which we should not run for clean up after successful run. For example cleaning up logs, cloning repo, merging PR branch. These tasks should run when deploying new metal3 dev env.
If we move these tasks out of cleanup then tags would not necessarily needed, I will try to move these tasks out of cleanup script.

jenkins/scripts/bare_metal_lab/tasks/cleanup-tasks.yaml
Sunnatillo: We cannot move them to deploy lab, because sequence is important here. What we can do create another playbook that includes tasks which are tagged as "deploy".

Sunnatillo: We clone metal3-dev-env to two different directories depending what we are testing. we need to run make clean from that directory as an essential part of cleanup. So, we cannot move these tasks to deploy-lab.yaml. But during cleanup after successful run, we need to somehow skip cloning and merging metal3-dev-env tasks again.

Sunnatillo: Other two tasks related finding and deleting log files also part of cleanup but we do not need to clean them after successful run. We might need the logs later for debuging or some other purposes. These tasks are not part of deploy tasks because of what they do.

lentzi90 (commented):
/lgtm

lentzi90 (commented):
/lgtm

furkatgofurov7 (approved):

lentzi90 (commented):
/lgtm

PR #441, opened 2022-10-06
Fix local image building in tests

Rozzii:


jenkins/scripts/integration_test_env.sh
```diff
@@ -13,26 +13,24 @@ then
   # for CAPM3 to verify the process (not BMO due to the build time for BMO image)
 
   if [[ "${UPDATED_BRANCH}" == "${REPO_BRANCH}" ]] && [[ "${UPDATED_REPO}" == *"${REPO_ORG}/${REPO_NAME}"* ]]; then
-    export BAREMETAL_OPERATOR_LOCAL_IMAGE="https://github.com/metal3-io/baremetal-operator.git"
-    export CAPM3_LOCAL_IMAGE="https://github.com/metal3-io/cluster-api-provider-metal3.git"
-    export CAPM3_LOCAL_IMAGE_BRANCH="${CAPM3RELEASEBRANCH}"
-    export BMOBRANCH="main"
```
lentzi90, line 19: I think we should still set `BMOBRANCH` here. From what I could find in metal3-dev-env, the `BMOBRANCH` is by default set to `BMORELEASE`. See https://github.com/metal3-io/metal3-dev-env/blob/5c29964fedbe3e9c8b47b16d537495787df6c409/lib/releases.sh#L49

Rozzii: yep you are right

lentzi90: I don't understand. Why would there not be a release?

lentzi90 (commented):
/lgtm

kashifest (approved):

PR #444, opened 2022-11-09
Fix BMOBRANCH=main only for main periodic test

kashifest:
This PR also noticed that the comment that we are not building BMO image is incorrect since we were. So since we were building CAPM3 and BMO images already for peiodic tests, this PR also adds the option to build IPAM image locally for periodic tests.


lentzi90 (commented):
/lgtm

furkatgofurov7 (approved):

PR #448, opened 2022-11-16
Fix BMOBRANCH export

kashifest:
We were wrongly checking against REPO_BRANCH var since REPO_BRANCH refers to dev-env branch not the branch of CAPM3. We should check against CAPM3RELEASEBRANCH to make sure if the periodic test is running against main branch or release branch


smoshiur1237 (approved):
/lgtm

furkatgofurov7 (approved):

PR #452, opened 2022-12-08
Add release-1.3 CAPM3/IPAM branch configurations and notes

kashifest:


furkatgofurov7 (approved):

prow/config/config.yaml
```diff
@@ -368,6 +374,7 @@ presubmits:
   - name: gomod
     branches:
     - main
+    - release-1.3
```
lentzi90, line 377: Why do we add it here? Or maybe better question, why wasn't release-1.2 here before?

kashifest: release-1.2 is one block down, its in separate block because of golang docker image version

lentzi90: Ah that explains it! Thanks!

PR #455, opened 2022-12-13
prow: pin shellcheck to v0.8.0

tuminoid:
Shellcheck v0.9.0 released today broke all the shellchecks. Relying on a stable/latest tags expose us to this kind of unexpected breakage so pinning down shellcheck image to v0.8.0 with a digest.

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -295,7 +295,7 @@ presubmits:
         env:
         - name: IS_CONTAINER
           value: "TRUE"
-        image: docker.io/koalaman/shellcheck-alpine:stable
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/lgtm

lentzi90:
I think github/jenkins is having some trouble at the moment that is causing the status of the job to not show correctly here on the PR. Please don't trigger again before checking carefully if it actually failed. I made this mistake myself also.
As seen here, it only failed once. Then we retriggered because it looked like it was still failed or failed again, even though the new job was running.
image

kashifest (approved):

PR #459, opened 2022-12-29
test bml integration test with centos9 image

Sunnatillo:


Sunnatillo:
bml test is working with centos 9, this PR served its purpose

PR #461, opened 2023-01-03
Add prow test for Nordix/metal3-dev-tools

kashifest:
As a first step this PR enables shellcheck prow test.

### DIFF
--- a/prow/config/config.yaml
+++ b/prow/config/config.yaml
@@ -52,6 +52,7 @@ tide:
         - metal3-io/project-infra
         - metal3-io/static-ip-manager-image
         - metal3-io/ip-address-manager
+        - Nordix/metal3-dev-tools
       labels:
         - lgtm
         - approved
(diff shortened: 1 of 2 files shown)


prow/config/config.yaml
```diff
@@ -232,6 +233,21 @@ periodics:
             secretName: github-token
 
 presubmits:
+  Nordix/metal3-dev-tools:
+  - name: shellcheck
```
tuminoid, line 237: yaml lists need to be intended by two spaces. Please indent all lists (lines starting with `-`) by two from previous indent level.

tuminoid (changes requested):

prow/config/config.yaml
kashifest: so whats our yamllint is checking then?

tuminoid: It is only configured to check for trailing spaces, duplicate keys, and truthy values = it doesn't, the config sucks. But the next guy who wants to change the yaml who has yaml formatter configured in their editor will have huge diff.

Vscode has nice yaml plugin from RH you should absolutely install if you work with yaml files btw.

kashifest: Thanks, done

kashifest: I will push changes to yamllint workflow.

prow/config/plugins.yaml
```diff
@@ -5,6 +5,29 @@ owners:
   skip_collaborators:
   - "metal3-io"
 plugins:
+  Nordix:
+    plugins:
+    - approve
```
tuminoid, line 10: This file would also need yaml formatter.

kashifest: Done

tuminoid (approved):
lgtm, but I'm no prow expert, leaving further review for others

lentzi90 (commented):
/lgtm

furkatgofurov7 (approved):

PR #472, opened 2023-01-18
Move ephemeral cluster from jumphost to internal server

Sunnatillo:
This PR moves running ephemeral cluster from jumphost to internal server in bml. 
1. Jumphost is critical for our lab. Dependencies of project always change and require newer package versions. It creates a risk of breaking OS at any point. OS that we test in CI already newer version than in jumphost. We have already seen package conflicts several times. 
2. After changes Ephemeral cluster (minikube) runs on internal server. We still use jumphost as a proxy to reach internal server.

### DIFF
--- a/jenkins/scripts/bml_integration_test.sh
+++ b/jenkins/scripts/bml_integration_test.sh
@@ -42,7 +42,8 @@ GINKGO_SKIP="${GINKGO_SKIP:-}"
 EPHEMERAL_TEST="${EPHEMERAL_TEST:-false}"
 BARE_METAL_LAB=true
 
-TEST_EXECUTER_IP="129.192.80.20"
+JUMPHOST_IP="129.192.80.20"
+TEST_EXECUTER_IP="192.168.1.3"
 
 cat <<-EOF > "${CI_DIR}/files/vars.sh"
 REPO_ORG="${REPO_ORG}"
@@ -65,11 +66,18 @@ EOF
 
 cat "${CI_DIR}/integration_test_env.sh" >> "${CI_DIR}/files/vars.sh"
 
+declare -a SSH_OPTIONS=(
+    -o StrictHostKeyChecking=no
+    -o UserKnownHostsFile=/dev/null
+    -o ServerAliveInterval=15 
+    -o ServerAliveCountMax=10
+    -i "${METAL3_CI_USER_KEY}"
+    -o ProxyCommand="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i ${METAL3_CI_USER_KEY} -W %h:%p ${METAL3_CI_USER}@${JUMPHOST_IP}"
+)
+
 # Send Remote script to Executer
 scp \
-  -o StrictHostKeyChecking=no \
-  -o UserKnownHostsFile=/dev/null \
-  -i "${METAL3_CI_USER_KEY}" \
+  "${SSH_OPTIONS[@]}" \
   -r \
   "${CI_DIR}/files/run_integration_tests.sh" \
   "${CI_DIR}/files/vars.sh" \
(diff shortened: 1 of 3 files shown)


tuminoid (commented):
lgtm, but someone with more bml knowledge should actually lgtm this

furkatgofurov7 (commented):
Changes look fine to me, however I am not fully aware of the 2 things with this PR, so if you could add/expand it in PR description, it would be very helpful:
1. Why do we need this change and what is the benefit? 
2. What would be different after this lands? Do we still keep using jumphost as a proxy between the user and new internal server?

Sunnatillo:
> Changes look fine to me, however I am not fully aware of the 2 things with this PR, so if you could add/expand it in PR description, it would be very helpful:
> 
>     1. Why do we need this change and what is the benefit?
> 
>     2. What would be different after this lands? Do we still keep using jumphost as a proxy between the user and new internal server?

Updated

jenkins/scripts/bml_integration_test.sh
```diff
@@ -43,6 +43,7 @@ EPHEMERAL_TEST="${EPHEMERAL_TEST:-false}"
 BARE_METAL_LAB=true
 
 TEST_EXECUTER_IP="129.192.80.20"
+INTERNAL_SERVER_IP="192.168.1.3"
```
furkatgofurov7, line 46: I think we need to make sure that file description on the script usage in the beginning of this file is also up-to-date with the current changes (GitHub does not allow me to comment directly on that lines)

furkatgofurov7 (commented):
Have not checked the log fetching myself for the BML run, but is it working properly? 
I don't see it being changed with current PR but we are specifying `TEST_EXECUTER_IP` in https://github.com/metal3-io/project-infra/blob/1285ef122e18ac40309e49505f60e71cf6536707/jenkins/scripts/fetch_logs.sh#L37

jenkins/scripts/bml_integration_test.sh
Sunnatillo: Good catch. I will add the changes for fetching logs script as well

jenkins/scripts/fetch_logs.sh
```diff
@@ -46,31 +47,66 @@ else
...
-  -o ServerAliveInterval=15 \
-  -i "${METAL3_CI_USER_KEY}" \
-  "${METAL3_CI_USER}"@"${TEST_EXECUTER_IP}" \
-  PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin \
-  /tmp/run_fetch_logs.sh "logs-${BUILD_TAG}.tgz" \
-  "logs-${BUILD_TAG}" "${IMAGE_OS}" "${TESTS_FOR}"
+if [ "${BARE_METAL_LAB}" == true ]; then
+  # Send Remote script to BML server
+  scp \
+    -o StrictHostKeyChecking=no \
```
tuminoid, line 54: This is excessively repetitive and error-prone way to call ssh/scp.

I'd suggest defining the common options something like this:

```bash
declare -a SSH_OPTIONS=(
    -o StrictHostKeyChecking=no
    -o UserKnownHostsFile=/dev/null
    -o ServerAliveInterval=15
    -i "${METAL3_CI_USER_KEY}"
    -o ProxyCommand="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i ${METAL3_CI_USER_KEY} -W %h:%p ${METAL3_CI_USER}@${TEST_EXECUTER_IP}"
)
```
and then all the calls would looks like
```bash
ssh \
  "${SSH_OPTIONS[@]}" \
  <user>@<port> \
  <whatever command>
```

jenkins/scripts/bml_integration_test.sh
```diff
@@ -70,11 +71,12 @@ scp \
   -o StrictHostKeyChecking=no \
   -o UserKnownHostsFile=/dev/null \
   -i "${METAL3_CI_USER_KEY}" \
+  -o "ProxyCommand ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ${METAL3_CI_USER_KEY}  ${METAL3_CI_USER}@${TEST_EXECUTER_IP} -W %h:%p" \
```
tuminoid, line 74: Same comment applies here.

tuminoid (changes requested):

jenkins/scripts/fetch_logs.sh
Sunnatillo: Thank you for pointing it out.
Changed it according your suggestion.

jenkins/scripts/bml_integration_test.sh
Sunnatillo: done

tuminoid (approved):
/lgtm

jenkins/scripts/bml_integration_test.sh
furkatgofurov7: Looks like logs collected properly now

furkatgofurov7 (approved):
/approve
/hold leaving it to @kashifest to take a second look and unhold if looks good.

kashifest (commented):
/hold cancel

PR #475, opened 2023-01-30
🌱 Add e2e integration as required checks

mboukhalfa:


prow/config/config.yaml
```diff
@@ -157,6 +165,7 @@ branch-protection:
               required_status_checks:
                 contexts:
                   [
+                    "test-centos-integration-e2e-main",
                     "test-centos-integration-release-1-3",
                     "test-ubuntu-integration-main",
```
lentzi90, line 170: Do we need 3 required checks here? I'm thinking that it would be enough to replace the ubuntu integration test.
```suggestion
                    "test-centos-integration-release-1-3",
                    "test-ubuntu-integration-e2e-main",
```

mboukhalfa: I wanted to keep Ansible main tests for sometime before removing them and we still do not have integration e2e for releases

lentzi90: Not sure I understand. This would still run the normal integration test on the release branch.
For `baremetal-operator` on line 108 we do almost the same: keep one normal integration test and replace the other with e2e integration test.

mboukhalfa: On BMO we do not require a test on releases so only main tests do you think here as well test on release-1.3 is not required

lentzi90: Ok, but here we actually add the e2e and keep both old integration tests. That was my point. Can we not just replace `test-ubuntu-instegration-main` with `test-ubuntu-integration-e2e-main`? We still run the integration test for the release branch

mboukhalfa: then I add Ubuntu e2e and remove the ansible one

lentzi90 (commented):
/lgtm

furkatgofurov7 (commented):
How long does it take to run the new e2e integration tests?

We kept a single mandatory job for some repos (i.e CAPM3, IPAM) because of resource constraints in the past, but with this PR it'll be changed and we need to run 2 jobs instead.  I do not know if we have enough quotas in CI for that

mboukhalfa:
> How long does it take to run the new e2e integration tests?
> 
> We kept a single mandatory job for some repos (i.e CAPM3, IPAM) because of resource constraints in the past, but with this PR it'll be changed and we need to run 2 jobs instead. I do not know if we have enough quotas in CI for that

This is mainly replace Ansible mandatory tests with equivalent e2e as a step towards deprecating the ansible tests So this will not have any impact on the quota (e2e integration takes around 30 min same as ansible)

furkatgofurov7:
> So this will not have any impact on the quota

how is it so, if you are running 2 mandatory tests in every PR instead of 1?

mboukhalfa:
> > So this will not have any impact on the quota
> 
> how is it so, if you are running 2 mandatory tests in every PR instead of 1?

Emm yes for cpm3 and ipam ! there were only 1 ansible test so instead of replacing it directly I thought to add e2e for sometime then remove the Ansible afterwards

furkatgofurov7:
> Emm yes for cpm3 and ipam ! there were only 1 ansible test so instead of replacing it directly I thought to add e2e for sometime then remove the Ansible afterwards

it is only in main and not in release branches, so for now should be good to go. 

/approve

PR #487, opened 2023-02-19
Fix serial port log collection in bml

Sunnatillo:
ssh-rsa HostKeyAlgorithms is not enabled by default in current ssh version. This PR fixes log collection from virtual serial port of servers in bml


lentzi90 (commented):
/approve

lentzi90:
I see approves from both me an Furkat, setting lgtm
/lgtm

PR #491, opened 2023-03-08
drop nonexisting check for mariadb-image

tuminoid:
Tide can't merge anything there as tide keeps waiting for this, but any steps to run it do not exist.


dtantsur (approved):

tuminoid:
/close

Test has been broken forever, but now it is fixed in Jenkins, so this PR is not needed.

PR #492, opened 2023-03-09
review approval shall not act as /approve

tuminoid:
Approving a review has acted same as explicit /approve if you have approval rights in the OWNERS. Let's make it so that if you want to approve a PR, you need to be explicit about it.

This is really badly documented in PROW, but this comment is the key: https://github.com/kubernetes/test-infra/blob/adf0d415167f1cba4cff708033f4184ad8cf3f33/prow/plugins/approve/approve.go#LL367C49-L367C68


tuminoid:
`review_acts_as_approve` is only found in that comment, there is no option to actually handle that config. Sad :(

/close

PR #500, opened 2023-04-03
add shellcheck and markdownlint to mariadb-image

tuminoid:
Add shellcheck and markdownlint checks for mariadb-image to prow.


kashifest (approved):
I would rather suggest to take this PR in first. There are only two PRs currently in mariadb, one which adds the two scripts and one on hold. Once this merges , it will run the jobs on your PR and we can know if things are broken regarding linters in main branch and push the fixes in the same PR so that main branch is not broken.

lentzi90 (commented):
/approve

lentzi90:
Ah already approved. I'm putting lgtm then
/lgtm

PR #504, opened 2023-04-06
Fix unsetting NUM_NODES variable and passing it to BML

Sunnatillo:
This PR:
Sets NUM_NODES as  env variable, instead of passing it trough vars.sh file. This makes it possible to set different value during configure host ansible task, where variable should be zero to avoid creating libvirt VMs (as real bare metal servers are used).
Fixes unsetting  NUM_NODES variable since the changes here: https://github.com/metal3-io/metal3-dev-env/pull/1193


smoshiur1237 (approved):
/lgtm

jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -30,7 +30,7 @@ then
   export KUBERNETES_VERSION="v1.23.8"
 fi
 
-if [ "${NUM_NODES}" == "null" ]
+if [ -z "${NUM_NODES+set}" ] | [ "${NUM_NODES}" == "null" ];
```
tuminoid, line 33: There should be `||` (or) instead of `|` (pipe), and I don’t follow what you’re trying to do with `+set`?

tuminoid (changes requested):

jenkins/scripts/files/run_integration_tests.sh
Sunnatillo: my bad, Fixed it.

jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -12,11 +12,13 @@ export CAPM3_VERSION
...
 export IMAGE_OS
 export DEFAULT_HOSTS_MEMORY
-export NUM_NODES
 export EPHEMERAL_TEST
 export GINKGO_FOCUS
 export GINKGO_SKIP
 export SKIP_CLEANUP
+# unsetting NUM_NODES when it is unbound
+# in BML tests it is not passed through vars file
+export NUM_NODES="$NUM_NODES:-"
```
tuminoid, line 21: You need to have the braces here, `export NUM_NODES="${NUM_NODES:-}"`

tuminoid (changes requested):

jenkins/scripts/files/run_integration_tests.sh
Sunnatillo: thanks. Added it.

tuminoid (approved):
/lgtm

Sunnatillo:
/test-integration-bml-centos

Previous run failed with a flake, not related to this change.

lentzi90 (commented):
/approve

PR #506, opened 2023-04-11
Set default value for IRONIC_USE_MARIADB

lentzi90:
Jobs that doesn't have the variable set will otherwise fail due to unset variable. The default is set to "false".


smoshiur1237 (approved):
lgtm

kashifest (approved):

lentzi90:
Keepalived is failing because of an unrelated issue.

```
/bin/manage-keepalived.sh: line 11: /usr/sbin/keepalived: Operation not permitted
```

PR #507, opened 2023-04-12
Remove BMOBRANCH export

kashifest:
Removing dictating BMOBRANCH var from project-infra. This was set here to only set main branch periodic tests with main BMO branch and the PR jobs with release tag. But now, we cant afford to do that anymore since main branch BMO is nonbackwards compatible. So dev-env dictates which branch of BMO will be tested. Fixing that in the PR https://github.com/metal3-io/metal3-dev-env/pull/1208


tuminoid (commented):
/lgtm

lentzi90 (commented):
/approve

PR #508, opened 2023-04-13
New Prow instance and docs

lentzi90:
The Metal3 Prow instance has been replaced with a new that runs in a Cluster API cluster on OpenStack (Cleura). It is using the S3 API in Cleura for object storage (instead of MinIO in front of Azure for the old instance).

This commit also includes basic docs for how to set up the cluster and deploy Prow.

Main changes:

- Run Prow in a CAPI/CAPO cluster instead of hand crafted kubeadm cluster
- Cinder volume for ghproxy instead of host path on the node
- HTTP01 challenge for certificates instead of DNS01 since we no longer have direct API access for DNS management.
- Removed old ingress subdomain that was unused (hook.apps.test.metal3.io).
- S3 storage in Cleura instead of MinIO in front of Azure
- New kustomizations for Prow, "infra" and the cluster itself.
- Moved label sync CronJob to the prow namespace (previously default).
- Changes in folder/file structure. Some files have been moved without changes.

TODO:

- [x] Docs for setting up GitHub bots
- [x] Docs for cherry-picker token
- [x] Update config and plugin paths to match the new folder structure.
- [x] Update Makefile


lentzi90:
Hmm now that I changed the paths in the config, it reformatted the files automatically and caused huge diffs. I put it as a separate commit for easier review but if you prefer I can skip the reformat and just change the paths.

tuminoid:
> Hmm now that I changed the paths in the config, it reformatted the files automatically and caused huge diffs. I put it as a separate commit for easier review but if you prefer I can skip the reformat and just change the paths.

If you refer to Prow config.yaml and plugins.yaml, we need to settle first are they supposed to be vanilla-YAML or k8s-yaml, and then format accordingly. Since Prow is coming from k8s, I believe the correct formatting would be k8s-YAML and in that case it would be good to do another PR just to correct the formatting, to keep the diff down. This is rather massive PR to begin with, and this would make it easier to review.

lentzi90:
Reverted the config files formatting now. I have also added the missing pieces to the readme.

lentzi90:
Oh, line length...
Also, the job that checks the prow config is unhappy since it cannot find the config files where it expects them. To do this properly I would need to put copies of the config files back in their original places (with updated paths) and then do a follow up that removes them.

prow/README.md
```diff
@@ -4,206 +4,383 @@ Metal3 Prow dashboard: <https://prow.apps.test.metal3.io>
...
-       --set controller.hostNetwork=true,controller.service.type=""\
-   ,controller.kind=DaemonSet --generate-name
-   helm repo update
+   ```bash
+   s3cmd --config .s3cfg --configure
+   # Provide Access key and Secret key from the openstack credentials created above.
+   # Leave Default Region as US.
+   # Set S3 Endpoint to s3-kna1.citycloud.com:8080
+   # Set DNS-style bucket+hostname:port template to the same s3-kna1.citycloud.com:8080
+   # Default values for the rest. And save settings.
```
kashifest, line 274: This section needs a bit more explanation. What is this used for?

prow/README.md
```diff
@@ -4,206 +4,383 @@ Metal3 Prow dashboard: <https://prow.apps.test.metal3.io>
...
-   ```shell
-   helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx
-   helm install ingress-nginx/ingress-nginx \
-       --set controller.hostNetwork=true,controller.service.type=""\
-   ,controller.kind=DaemonSet --generate-name
-   helm repo update
+   ```bash
+   s3cmd --config .s3cfg --configure
+   # Provide Access key and Secret key from the openstack credentials created above.
+   # Leave Default Region as US.
```
kashifest, line 271: US?

prow/infra/service.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: ingress-nginx-controller
+  namespace: ingress-nginx
+  annotations:
+    loadbalancer.openstack.org/keep-floatingip: "true"
+spec:
+  loadBalancerIP: 188.95.227.72
```
kashifest, line 9: can we set it as envvar and not put it in git?

prow/README.md
```diff
@@ -4,206 +4,383 @@ Metal3 Prow dashboard: <https://prow.apps.test.metal3.io>
...
+   "networks": "bda060e0-07a2-4ba2-8fe0-dba7626db240",
+   "flavor": "4C-4GB-20GB",
+   "floating_ip_network": "ext-net",
+   "ssh_username": "ubuntu",
+   "volume_type": "",
+   "kubernetes_deb_version": "1.26.3-00",
+   "kubernetes_rpm_version": "1.26.3-0",
+   "kubernetes_semver": "v1.26.3",
+   "kubernetes_series": "v1.26"
+}
```
kashifest, line 108: so uplifting k8s version would mean creating node images and doing rolling upgrades? The images stay in openstack only right?

kashifest (approved):
Nice work! Few nits/questions inline

prow/README.md
lentzi90: Yeah I don't get why, but it must be US or it won't work.

prow/infra/service.yaml
lentzi90: Why? It would require a templating step to make the manifests usable then. It is not like this IP is secret, anyone can find it with `host prow.apps.test.metal3.io`.

kashifest: I was thinking if adding a publicly reachable IP in the manifest in git increases attack surface? But yeah its also true that it is not something that is a secret after all.

prow/README.md
lentzi90: That said, we still specify the region as part of the URL and I guess that is what matters for Cleura.

prow/README.md
lentzi90: Good point! I'll add that it is used for creating the buckets.

lentzi90: If it is ok, I could do this in a follow up where I also remove the extra copies of config that is now there just to make the config-validator happy.

kashifest: Yeah thats fine.

prow/README.md
```diff
@@ -4,206 +4,383 @@ Metal3 Prow dashboard: <https://prow.apps.test.metal3.io>
...
+1. Create a personal access token for the GitHub bot account.
+   This should be done from the [metal3-io-bot](https://github.com/metal3-io-bot)
+   GitHub bot account.
+   You can follow this [link](https://github.com/settings/tokens) to create the
+   token. When generating the token, make sure you have only the following scopes
+   checked in.
 
-1. Create HMAC token for webhook validation and generate a secret out of it.
-   Don't forget to add this token on GitHub webhook configuration.
+   - `repo` scope for full control of private repositories
```
tuminoid, line 164: Instead of private, maybe say `all` repositories. I know its Github wording, but I found it annoying when its just implicitly including public ones.

prow/README.md
```diff
@@ -4,206 +4,383 @@ Metal3 Prow dashboard: <https://prow.apps.test.metal3.io>
...
+   - `workflow`
+   - `admin:org_hook` for a github org
```
tuminoid, line 166: I wonder if we really need this permission.

prow/README.md
```diff
@@ -4,206 +4,383 @@ Metal3 Prow dashboard: <https://prow.apps.test.metal3.io>
...
+- `HMAC_TOKEN`
+- `GITHUB_TOKEN`
+- `CHERRYPICK_TOKEN`
+
+Now you are ready to create the files.
+
+1. Create `clouds.yaml` secret used by the CAPO controllers to manage the infrastructure.
+
+   ```bash
+   cat > capo-cluster/clouds.yaml <<-EOF
```
tuminoid, line 220: `-EOF` is removing leading tabs, which you don't use here, so remove the `-`, it does nothing here. It would be wrong to remove leading spaces anyways.

tuminoid (commented):
Some nits, but overall great work @lentzi90 !

prow/README.md
kashifest: No wait it says use it `if you plan on handling a github org`, our prow is not handling anything for org right? Only thing is webhook which is deployed for org.

lentzi90:
Alright, I think I have addressed all comments. I have not squashed the 3 commits though. They make it quite a lot easier to understand what happened to the config. But if you want to have it all squashed, let me know and I will do it.

prow/README.md
lentzi90: Added it now when I anyway got more comments to address :slightly_smiling_face:

PR #514, opened 2023-04-28
Add e2e integration as required job for release-1.4 branch

kashifest:
Gradually replacing integration tests with e2e-integration to be the required test. At this step we do that for release-1.4 branch.

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -130,7 +130,7 @@ branch-protection:
                 contexts: ["test-ubuntu-integration-release-1-3"]
             release-1.4:
               required_status_checks:
-                contexts: ["test-ubuntu-integration-release-1-4"]
+                contexts: ["test-ubuntu-e2e-integration-release-1-4"]
         ironic-image:
           branches:
             main:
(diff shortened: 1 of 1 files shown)


kashifest:
> > _ No description provided. _
> 
> let's tell the whole story

Done

prow/manifests/overlays/metal3/config.yaml
```diff
@@ -161,7 +161,7 @@ branch-protection:
                 contexts: ["test-ubuntu-integration-release-1-3"]
             release-1.4:
               required_status_checks:
-                contexts: ["test-ubuntu-integration-release-1-4"]
+                contexts: ["test-ubuntu-e2e-integration-release-1-4"]
```
mboukhalfa, line 164: same for line 177 no more ansible tests on releases

kashifest: thats for dev-env, we should only do it for capm3 and ipam now I understood ? And this way we keep atleast the ansible scripts also being tested ?

mboukhalfa: My idea is to no loner support ansible tests on for the new releases but if you want to keep supporting ansible integrations on dev-env even when they are deprecated on capm3 we can keep it but we'll end up duplicating the same test with different implementations

smoshiur1237 (approved):
/lgtm

Sunnatillo (commented):
/lgtm

PR #518, opened 2023-05-08
Prow: Bump ingress-nginx to v1.7.1

lentzi90:
Also added TTL for the Jobs that comes with ingress-nginx. They are supposed to run each time we upgrade, but if the previous job is still there, then we get a naming collision.


tuminoid (commented):
/lgtm

Rozzii (approved):

kashifest (approved):

PR #521, opened 2023-05-11
Prow: PodDisruptionBudget for hook and deck

lentzi90:
Kubernetes upgrades can cause disruptions to hook and deck when the old node is drained. Both of them run with 2 replicas, but without a PodDisruptionBudget both replicas would anyway get evicted if they run one the same node. Adding PodDisruptionBudgets for them with minAvailable=1 ensures that only 1 replica is evicted at a time.

The other components of prow are not sensitive in the same way to disruptions. They will just catch up when they become available again. But hook and deck must be avaialable in order to receive requests.


kashifest (approved):

smoshiur1237 (commented):
/lgtm

PR #527, opened 2023-05-19
Prow: Upgrade k8s to v1.26.5

lentzi90:
Should not be merged before #520.

PR #528, opened 2023-05-19
Prow: add PodDisruptionBudget for ingress controller

lentzi90:
We already added PDBs for the prow components that are sensitive to disruptions. This adds PDB and scales the ingress controller to 2 replicas for the same reason. The trafic to the prow components go through the ingress controller so it must also be available for them to be reachable.

Should not be merged before #527


smoshiur1237 (approved):
/lgtm

kashifest (approved):

PR #530, opened 2023-05-23
Prow: Fix needs-rebase servie namespace

lentzi90:
The namespace for the needs-rebase service was wrong (default instead of prow). This resulted in webhook requests not reaching the needs-rebase plugin.
Also, the service had type NodePort for no good reason. This is now changed to ClusterIP.


tuminoid (commented):
/lgtm

Fix typo in title.

kashifest (approved):

PR #536, opened 2023-06-07
increase node_reuse timeout to 5 hours in parallel e2e

Rozzii:
Currently based on consistent test results of the parallel tests, node_reuse doesn't have enough time to fininsh all the test actions before it times out.


kashifest (approved):

lentzi90 (commented):
/lgtm

PR #540, opened 2023-06-15
Prow: Resources and LimitRanges

lentzi90:
There are two commits in this PR. The first commit is for upgrading kubernetes to v1.26.6, the second is for reserving resources across the cluster. More details below (from commit messages):

    This upgrades the Kubernetes cluster to v1.26.6 and includes a "safety
    handle" that was recently added in CAPO templates, namely to set the
    provider ID directly through the kubelet. This prevents potential issues
    from differences between the OpenStackMachine and server names. If they
    differ for some reason, then the cloud controller will not be able to
    set the correct provider ID. By telling the kubelet to set the provider
    ID directly, we bypass this completely.

    This commit adds reserved resources for all nodes. It also sets resource
    requests on prow job init containers and sidecar and adds a LimitRange
    to set default requests for all prow job containers.
    
    The reserved resources is to make sure that the kubelet and other
    critical resources on the nodes get enough resources to function
    properly. The LimitRange is to get some kind of default for all jobs. It
    will not be a good match for most of them, but until we have precise
    knowledge it is better to set something than leaving it empty. This will
    also ensure that if we in the future forget to set resource requests
    when adding new jobs, they will still get the defaults.
    
    Finally, the init container and sidecar requests are there to avoid
    setting the (quite large) default requests on these containers. All jobs
    have the side car and the init containers and they do not consume much
    resources. Unless we set something for them, they would also get the
    defaults.

### DIFF
--- a/prow/capo-cluster/kubeadmcontrolplane.yaml
+++ b/prow/capo-cluster/kubeadmcontrolplane.yaml
@@ -16,12 +16,16 @@ spec:
         kubeletExtraArgs:
           cloud-provider: external
           provider-id: "openstack:///'{{ instance_id }}'"
+          kube-reserved: cpu=200m,memory=100Mi
+          system-reserved: cpu=100m,memory=100Mi
         name: '{{ local_hostname }}'
     joinConfiguration:
       nodeRegistration:
         kubeletExtraArgs:
           cloud-provider: external
           provider-id: "openstack:///'{{ instance_id }}'"
+          kube-reserved: cpu=200m,memory=100Mi
+          system-reserved: cpu=100m,memory=100Mi
         name: '{{ local_hostname }}'
   machineTemplate:
     infrastructureRef:
--- a/prow/capo-cluster/kubeadmcontrolplane.yaml
+++ b/prow/capo-cluster/kubeadmcontrolplane.yaml
@@ -15,16 +15,18 @@ spec:
       nodeRegistration:
         kubeletExtraArgs:
           cloud-provider: external
+          provider-id: "openstack:///'{{ instance_id }}'"
         name: '{{ local_hostname }}'
     joinConfiguration:
       nodeRegistration:
         kubeletExtraArgs:
           cloud-provider: external
+          provider-id: "openstack:///'{{ instance_id }}'"
         name: '{{ local_hostname }}'
   machineTemplate:
     infrastructureRef:
       apiVersion: infrastructure.cluster.x-k8s.io/v1alpha6
       kind: OpenStackMachineTemplate
-      name: prow-control-plane-v1-26-5
+      name: prow-control-plane-v1-26-6
   replicas: 1
-  version: v1.26.5
+  version: v1.26.6
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -28,6 +28,23 @@ plank:
         entrypoint: gcr.io/k8s-prow/entrypoint:v20230329-c93d79fb7d
         initupload: gcr.io/k8s-prow/initupload:v20230329-c93d79fb7d
         sidecar: gcr.io/k8s-prow/sidecar:v20230329-c93d79fb7d
+      resources:
+        clonerefs:
+          requests:
+            cpu: 10m
+            memory: 10Mi
+        initupload:
+          requests:
+            cpu: 10m
+            memory: 10Mi
+        place_entrypoint:
+          requests:
+            cpu: 10m
+            memory: 10Mi
+        sidecar:
+          requests:
+            cpu: 10m
+            memory: 10Mi
 
 tide:
   merge_method:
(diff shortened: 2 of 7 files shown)


lentzi90:
If you want to learn more about reserving resources for kubelet and system daemons, this is a good resource: https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/

For the init containers and sidecar, I looked at what they have for the k/k prow cluster [here](https://github.com/kubernetes/test-infra/blob/79ded6a8d1f71270ad0b6f35c8738ac6c23eadb4/config/prow/config.yaml#L26-L38), but decided to lower the CPU request based on measured usage. The sidecar seems to use only 1m CPU. Init containers are so fast that I could not properly measure them, but they will anyway stop before the real test so they won't impact much.

prow/capo-cluster/kubeadmcontrolplane.yaml
```diff
@@ -15,16 +15,22 @@ spec:
...
+          provider-id: "openstack:///'{{ instance_id }}'"
+          kube-reserved: cpu=200m,memory=100Mi
+          system-reserved: cpu=100m,memory=100Mi
         name: '{{ local_hostname }}'
   machineTemplate:
     infrastructureRef:
       apiVersion: infrastructure.cluster.x-k8s.io/v1alpha6
       kind: OpenStackMachineTemplate
-      name: prow-control-plane-v1-26-5
+      name: prow-control-plane-v1-26-6
```
kashifest, line 34: Any specific reason for not going to next minor version i.e. v1.27.x?

prow/capo-cluster/kubeadmcontrolplane.yaml
```diff
@@ -15,16 +15,22 @@ spec:
       nodeRegistration:
         kubeletExtraArgs:
           cloud-provider: external
+          provider-id: "openstack:///'{{ instance_id }}'"
```
kashifest, line 18: humm? how was it working absent provider-id?

prow/manifests/overlays/metal3/config.yaml
```diff
@@ -28,6 +28,23 @@ plank:
...
+            cpu: 10m
+            memory: 10Mi
+        place_entrypoint:
+          requests:
+            cpu: 10m
+            memory: 10Mi
+        sidecar:
+          requests:
+            cpu: 10m
+            memory: 10Mi
```
kashifest, line 47: cant we set this as a default like you are doing for limits in limitrange.yaml ?

kashifest (approved):

prow/capo-cluster/kubeadmcontrolplane.yaml
kashifest: ok makes sense

prow/capo-cluster/kubeadmcontrolplane.yaml
lentzi90: When this is not specified, the cloud controller compares the names of the OpenStackMachines or Node with the names of the servers in the OpenStack API and sets the provider ID based on that.

We saw some issues in CAPO when names contains special characters, because OpenStack modifies the name automatically for the server but this is not done for Nodes and other resources. That way you can get mismatch and then the cloud controller is unable to set the provider ID. This is why we have now added it to avoid these issues.

prow/manifests/overlays/metal3/config.yaml
lentzi90: I'm not understanding. What do you mean by set it as default?
This is setting requests for the sidecar and the numbers are based on the actual usage I saw in the cluster (cpu 1m and memory 9Mi). If we don't set the request here, then it would get the default from the LimitRange (which is way too much for the sidecar)

lentzi90: The LimitRange is for all containers. This is for specific prow containers that are named exactly like here. So `sidecar` here is not just any sidecar it is a container that is named `sidecar` and runs together with every prow job. This is how it looks like:

```yaml
apiVersion: v1                                                                                                                                                                                             
kind: Pod                                                                                                                                                                                                  
metadata:                                                                                                                                                                                                                                                                                                                                                                        
  name: a178259f-0b77-11ee-b158-7230da3050f0                                                                                                                                                               
  namespace: test-pods                                                                                                                                                                                                                                                                                                                                                
spec:                                                                                                                                                                                                                                                                                                                                                                     
  containers:                                                                                                                                                                                              
  - name: test
    ...
    # Resources here would come from the LimitRange unless we specify something in the config
  - name: sidecar   # <------ This is the sidecar
    resources:
      requests:
        cpu: 10m
        memory: 10Mi
  initContainers:
  - name: clonerefs
    ...
  - name: initupload
    ...
  - name: place-entrypoint
    ...
```

kashifest: oh ok, my bad, thanks for explaining, I was under the impression it is type and we could set it for all like we are doing in limits.

Sunnatillo (commented):
/lgtm

PR #543, opened 2023-06-21
change the ipa-downloader required test to centos

tuminoid:
ipa-downloader is an init container, which is k8s concept. On Ubuntu integration, it is manually executed (using wrong image, currently), so it is not actually testing it properly. Centos one is running the init container as it should be running, so making it the mandatory test for ipa-downloader repo.


lentzi90 (commented):
Makes sense!
/approve

PR #544, opened 2023-06-22
Run golint and gofmt on more changes in BMO

lentzi90:
Golint and gofmt were configured to only run then .go files were changed, but these jobs could also be affected by changes to go.mod and other files (e.g. version changes).

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -321,7 +321,7 @@ presubmits:
             imagePullPolicy: Always
   metal3-io/baremetal-operator:
     - name: gofmt
-      run_if_changed: '\.go$'
+      skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
       decorate: true
       spec:
         containers:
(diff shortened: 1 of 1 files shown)


prow/manifests/overlays/metal3/config.yaml
```diff
@@ -321,7 +321,7 @@ presubmits:
             imagePullPolicy: Always
   metal3-io/baremetal-operator:
     - name: gofmt
-      run_if_changed: '\.go$'
+      run_if_changed: '(((^|/)OWNERS)|(\.md))$'
```
tuminoid, line 324: Shouldn’t it be `skip_if_only_changed`?

tuminoid (changes requested):

prow/manifests/overlays/metal3/config.yaml
lentzi90: Thanks! Fixed now

tuminoid (approved):

lentzi90:
Since Kashif is on vacation and this is such a small change that already got two approving reviews, I will add the approve label myself
/approve

PR #549, opened 2023-07-07
Fixes confusing log placements on failing runs.

huutomerkki:
Before_pivoting folder will only be created if there is a snapshot to put into it. After_pivoting folder renamed to final_logs.

### DIFF
--- a/jenkins/scripts/files/run_fetch_logs.sh
+++ b/jenkins/scripts/files/run_fetch_logs.sh
@@ -70,13 +70,15 @@ fetch_k8s_logs()
 # Fetch k8s logs
 fetch_k8s_logs "management_cluster" "/home/metal3ci/.kube/config"
 
-# Fetch Ironic containers logs before pivoting to the target cluster
-CONTAINER_LOGS_DIR="${LOGS_DIR}/${CONTAINER_RUNTIME}/before_pivoting"
-mkdir -p "${CONTAINER_LOGS_DIR}"
-cp -r /tmp/"${CONTAINER_RUNTIME}"/* "${CONTAINER_LOGS_DIR}"
+# Fetch Ironic containers logs before pivoting to the target cluster, if they exist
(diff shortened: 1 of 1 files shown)


mboukhalfa (approved):
/lgtm

jenkins/scripts/files/run_fetch_logs.sh
```diff
@@ -70,13 +70,15 @@ fetch_k8s_logs()
 # Fetch k8s logs
 fetch_k8s_logs "management_cluster" "/home/metal3ci/.kube/config"
 
-# Fetch Ironic containers logs before pivoting to the target cluster
-CONTAINER_LOGS_DIR="${LOGS_DIR}/${CONTAINER_RUNTIME}/before_pivoting"
-mkdir -p "${CONTAINER_LOGS_DIR}"
-cp -r /tmp/"${CONTAINER_RUNTIME}"/* "${CONTAINER_LOGS_DIR}"
+# Fetch Ironic containers logs before pivoting to the target cluster, if they exist
+if [ -d /tmp/"${CONTAINER_RUNTIME}" ] && [ "$(ls /tmp/"${CONTAINER_RUNTIME}"/)" ]; then
```
Rozzii, line 74: the second condition is a no-op, if the dir exists ls will return 0 (true) all the time.

Rozzii: Use [ -n "$(ls -A /tmp/"${CONTAINER_RUNTIME}"/)"]

Rozzii (changes requested):

jenkins/scripts/files/run_fetch_logs.sh
Rozzii: okay looks like [ '' ] is interpreted as false so it is fine

Rozzii (approved):

PR #550, opened 2023-07-14
Bump golang to v1.20 for ipam main branch prow tests

Sunnatillo:
As title

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -932,6 +932,21 @@ presubmits:
     - name: gomod
       branches:
         - main
+      skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
+      decorate: true
+      spec:
+        containers:
+          - args:
+              - ./hack/gomod.sh
+            command:
+              - sh
+            env:
+              - name: IS_CONTAINER
+                value: "TRUE"
+            image: docker.io/golang:1.20
+            imagePullPolicy: Always
+    - name: gomod
+      branches:
         - release-1.3
         - release-1.4
       skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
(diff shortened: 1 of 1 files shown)


smoshiur1237 (commented):
/lgtm

prow/manifests/overlays/metal3/config.yaml
```diff
@@ -932,6 +932,21 @@ presubmits:
...
+        containers:
+          - args:
+              - ./hack/gomod.sh
+            command:
+              - sh
+            env:
+              - name: IS_CONTAINER
+                value: "TRUE"
+            image: docker.io/golang:1.20
+            imagePullPolicy: Always
```
Rozzii, line 947: I think "Always" is a bit aggressive for imagePullPolicy, it think this should be pulled only if the image is not present

Rozzii (changes requested):

prow/manifests/overlays/metal3/config.yaml
Sunnatillo: Noted. But there might be another reason, why we set `imagePullPolicy` to ` Always` for all images. I can test it in separately.

Sunnatillo: changed to IfNotPresent. It will prevent downloading images for each prow test. Nice catch.

kashifest (approved):

prow/manifests/overlays/metal3/config.yaml
```diff
@@ -1143,6 +1217,21 @@ presubmits:
     - name: generate
       branches:
         - main
+      skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
```
mboukhalfa, line 1220: cannot we filter all .md files (even README )

Sunnatillo: Can we take this as it is for now? I will handle this and `imagePullPolicy:IfNotPresent` in a separate PR

mboukhalfa: NP

Rozzii (approved):

PR #558, opened 2023-07-31
Fix make target for dev-env PRs

kashifest:
we need to test 01_  script also for dev-env PRs which is currently skipped since we use ci_run make target. This PR correct  for the dev-env PRs to use make command which includes the script.


smoshiur1237 (commented):
/lgtm

Sunnatillo (approved):
/lgtm

lentzi90 (commented):
/approve

PR #560, opened 2023-08-04
Add describe pods output during log collection

smoshiur1237:
We are missing some logs during the failure in target log collection. Where pods are not running and need describe logs to debug. Because if pods are not running, no logs will be generated by it. So adding describe pods output will tell us the reason of the failure of the pods.


kashifest (approved):

kashifest:
/approve cancel
as it is not working as expected

smoshiur1237:
So the log collection in successful run starts after deprovisioning the target cluster. That is the reason that it was not getting any namespace of pods from target cluster. So this will work during the target cluster is available.
/unhold

kashifest:
/approve
In my opinion we have to redesign the log collection totally. Target cluster logs should be available irrespective of a passed or failed test.

PR #562, opened 2023-08-10
re-enable IPA download in tests

Rozzii:
In fact the download was never disabled, dev-env is downloading IPA based on a different logic. Exporting the IPA_DOWNLOAD_ENABLED=false variable is not needed until there is an established process of pinning the IPA version for CI test jobs.

This variable is also blocking  https://github.com/metal3-io/metal3-dev-env/pull/1272 .


tuminoid (commented):
/lgtm

smoshiur1237 (approved):
/lgtm

tuminoid (approved):

PR #569, opened 2023-08-30
Prow: Disable ghproxy pushgateway

lentzi90:
We are not running a pushgateway so there is no point trying to push any metrics to it. This was causing lots of error messages in the ghproxy logs.


kashifest (approved):

smoshiur1237 (approved):
/lgtm

PR #570, opened 2023-09-01
add missing default value to KEEP_TEST_ENV

Rozzii:
This commit:
  - Adds a default value to the KEEP_TEST_ENV variable that is used during log fetching. The missing default value was causing "unbound variable" error.


tuminoid (commented):
/approve

lentzi90 (commented):
/lgtm

PR #571, opened 2023-09-01
Fix unbound variables for dev tools integration tests

Sunnatillo:
This PR fixes unbound variables issue, when running dev-tools-integration-tests.

PR #572, opened 2023-09-01
add default value to GINKGO_FOCUS variable

Rozzii:
This commit:
  - Adds missing default value to GINKGO_FOCUS variable used during log fetching. Missing default value was causing "unbound variable" error.


tuminoid:
Already handled in https://github.com/metal3-io/project-infra/pull/571

tuminoid:
Instead, fix KEEP_TEST_ENV and GINKGO_FOCUS in integration_delete.sh, which also fails to these issues.

tuminoid (changes requested):
:arrow_up:

Rozzii:
/cancel
There is already an other fix that is bit more extensive so let's go with that.

PR #590, opened 2023-10-24
Prow: Fix commenter and checkconfig command

lentzi90:
We have prow jobs that fail with

```
{"component":"entrypoint","error":"could not start the process: fork/exec /app/robots/commenter/app.binary: no such file or directory","file":"k8s.io/test-infra/prow/entrypoint/run.go:84","func":"k8s.io/test-infra/prow/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2023-10-24T17:46:51Z"}
```

There was a change in upstream that we missed when bumping the image tags.
See https://github.com/kubernetes/test-infra/pull/25384/files

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -219,7 +219,7 @@ periodics:
       containers:
         - image: gcr.io/k8s-prow/commenter:v20231011-33fbc60185
           command:
-            - /app/robots/commenter/app.binary
+            - commenter
           args:
             - |-
               --query=org:metal3-io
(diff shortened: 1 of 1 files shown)


smoshiur1237 (approved):
/lgtm

smoshiur1237:
Good catch, how did we miss it.

lentzi90:
The check-prow-config that run on the PR that changed the image tags did not use the new image :sweat_smile: 
Then we have just not checked the periodic stale jobs. We should probably add some alerting for that

lentzi90:
I have applied this in the cluster now (that is why the second check-prow-config was successful)

lentzi90:
I'm going to self approve this since it is such a small change and prow is broken without it
/approve

PR #592, opened 2023-11-01
WIP add jump_host support static workers

Rozzii:
This commit adds the support for establishing ssh connection between a static worker and a test executor vm via an ssh jumphost.


jenkins/scripts/fetch_logs.sh
```diff
@@ -63,6 +64,15 @@ if [[ "${BARE_METAL_LAB}" == true ]]; then
         -i "${METAL3_CI_USER_KEY}"
         -o ProxyCommand="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i ${METAL3_CI_USER_KEY} -W %h:%p ${METAL3_CI_USER}@${JUMPHOST_IP}"
     )
+elif [[ -n "${SSH_JUMP_HOST}" ]]; then
+    declare -a SSH_OPTIONS=(
+        -o StrictHostKeyChecking=no
```
tuminoid, line 69: We are duplicating a lot of these options. Maybe declare the array with before the `if` with the common options and only add the case specific ones inside the `if`

jenkins/scripts/integration_test.sh
```diff
@@ -179,33 +180,64 @@ fi
...
-  -o StrictHostKeyChecking=no \
-  -o UserKnownHostsFile=/dev/null \
-  -i "${METAL3_CI_USER_KEY}" \
-  "${CI_DIR}/files/${TEMP_FILE_NAME}" \
-  "${METAL3_CI_USER}@${TEST_EXECUTER_IP}:/tmp/vars.sh" >/dev/null
-
+if [[ -n "${SSH_JUMP_HOST}" ]]; then
+    # Send Remote script to Executer
+    scp \
+      -o StrictHostKeyChecking=no \
```
tuminoid, line 186: Please use the `ssh_options` array here too, these `scp` are becoming unreadable and unmaintainable. We should construct the ssh_options, and not duplicate the actual `scp` lines at all.

jenkins/scripts/integration_test_clean.sh
```diff
@@ -51,11 +52,13 @@ if [[ "${OS_REGION_NAME}" == "Fra1" ]] || [[ "${OS_PROJECT_NAME}" == "dev2" ]];
     TEST_EXECUTER_IP="${FLOATING_IP}"
 fi
 
+if [[ -n "${SSH_JUMP_HOST}" ]]; then
```
tuminoid, line 55: In this if, you should only be adding the proxy line to `ssh_options` and not duplicate the actual commands.

jenkins/scripts/utils.sh
```diff
@@ -34,35 +34,53 @@ wait_for_ssh()
...
     echo "Waiting for SSH connection to Host[${SERVER}]"
-    until ssh -o ConnectTimeout=2 \
-        -o StrictHostKeyChecking=no \
-        -o UserKnownHostsFile=/dev/null \
-        -i "${KEY}" \
-        "${USER}"@"${SERVER}" echo "SSH to host is up" > /dev/null 2>&1; do
-        sleep 1
-    done
-
+    if [[ -n "${JUMP_HOST}" ]]; then
```
tuminoid, line 40: Same as above. Do not duplicate the actual code.

jenkins/scripts/utils.sh
```diff
@@ -34,35 +34,53 @@ wait_for_ssh()
...
 
     USER="${1:?}"
     KEY="${2:?}"
     SERVER="${3:?}"
+    SSH_JUMP_HOST="${4:-}"
 
-    cloud_init_status=$(ssh -o ConnectTimeout=2 -o StrictHostKeyChecking=no \
-        -o UserKnownHostsFile=/dev/null -i "${KEY}" \
-        "${USER}"@"${SERVER}" cloud-init status --long --wait)
+    if [[ -n "${SSH_JUMP_HOST}" ]]; then
```
tuminoid, line 75: Same here.

tuminoid (commented):
Let's make this more maintainable.

Rozzii:
This is something that we don't really need at the moment, we have decided to move towards the dynamic workers for now.

PR #594, opened 2023-11-07
BML: prune docker storage

lentzi90:
We already prune containers, images and volumes, but docker was still taking up more than 60 GB. It was all cleaned up with docker system prune.


smoshiur1237 (approved):
/lgtm

mboukhalfa (approved):

PR #598, opened 2023-11-15
[debug] run ssh in tmux

mboukhalfa:


jenkins/scripts/integration_test.sh
```diff
@@ -200,6 +200,7 @@ rm -f "${CI_DIR}/files/${TEMP_FILE_NAME}"
 echo "Running the tests"
 # Execute remote script
 # shellcheck disable=SC2029
+tmux
```
tuminoid, line 203: I don't think you can use tmux in a script like this.

First of all, it needs interactive terminal. You can try via ssh `ssh <box> tmux` will fail with `open terminal failed: not a terminal`. Secondly, passing commands to tmux (or anything) cannot be input from a script like that.

See this: https://how-to.dev/how-to-create-tmux-session-with-a-script

mboukhalfa: I guess even if managed to run ssh command in tmux jenkins will have problem with showing the logs

PR #600, opened 2023-12-01
run markdownlint when changing the linter

tuminoid:
When bumping markdownlint I noticed the linter change isn't actually tested in Prow as no .md files were changed.

Run markdownlint when touching the linter script, so the rule ignores as well as component bumps get tested.


tuminoid:
Not tested by integration.

Sunnatillo (commented):
LGTM

kashifest (commented):
why aren't we bumping the markdown container images also here?

tuminoid:
> why aren't we bumping the markdown container images also here?

They are not all merged yet, and since this is not merged, they're not really tested.

kashifest:
> They are not all merged yet, and since this is not merged, they're not really tested.

/approve

PR #607, opened 2023-12-18
WIP: unpack make commands into script calls

tuminoid:
Test only, do not merge.


tuminoid:
Test actually did nothing, since the Ironic-image main is broken on Ubuntu right now...

PR #608, opened 2023-12-20
Add jenkins-operator for prow

lentzi90:
The name is a bit misleading unfortunately. This prow component is used to trigger jenkins jobs. It will allow us to have one common way to trigger all tests (except github workflows) and we can stop using the ghprb jenkins plugin that is no longer maintained.

### DIFF
--- a/prow/manifests/overlays/metal3/patches/jenkins-operator.yaml
+++ b/prow/manifests/overlays/metal3/patches/jenkins-operator.yaml
@@ -1,0 +1,14 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: prow
+  name: jenkins-operator
+spec:
+  template:
+    spec:
+      containers:
+      - name: jenkins-operator
+        resources:
+          requests:
+            cpu: 50m
+            memory: 50Mi
--- a/prow/.gitignore
+++ b/prow/.gitignore
@@ -7,3 +7,4 @@ service-account.json
 github-token
 cherrypick-bot-github-token
 hmac-token
+jenkins-token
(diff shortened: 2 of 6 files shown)


prow/manifests/overlays/metal3/patches/jenkins-operator.yaml
lentzi90, line 1: This may be silly of me, but we have done all other resource requests as a separate patch so I did the same here also.

lentzi90:
Not relevant

tuminoid:
Looks nice! How do we plan to test it?

lentzi90:
The jenkins-operator has already been installed in the cluster, but the config change has not been applied. That will happen when we merge this.
To test these changes I have applied the jenkins-operator manifests manually and configured it to use a separate configmap instead of the auto-synced. Then I created a prow job and applied it to see that the jenkins-operator was able to trigger the jenkins job. This worked well.

The part that I could not test is the deck integration that should allow it to display logs from jenkins. However, I have confirmed that it is possible to get the logs by curling the jenkins-operator pod directly so I'm relatively sure that it will work.

After we merge this, we will still need to add trigger configurations for all jobs that we want to trigger in this way. I think we should do that separately though, and perhaps not all jobs at once. We could for example start with the BMO e2e jobs that are currently not required as a low risk first step.

tuminoid:
Sounds fair! Timing just before holiday vacations though. It sounds like we can merge this, but maybe not mess with JJB and triggers just yet?

/approve

lentzi90:
Yeah that makes sense!

smoshiur1237 (approved):
/lgtm

PR #610, opened 2023-12-21
BMO e2e: Allow flexible artifact naming

lentzi90:
This is needed before we can add a matrix to the pipeline. Otherwise the parallel tests would overwrite each others artifacts...


smoshiur1237 (approved):
/lgtm

tuminoid (commented):
/approve

PR #614, opened 2024-01-09
enable build timestamps

Rozzii:
This commit enables build log timestamps in all pipeline scripts with a 1 second resolution.

### DIFF
--- a/jenkins/jobs/bmo_e2e_tests.pipeline
+++ b/jenkins/jobs/bmo_e2e_tests.pipeline
@@ -44,7 +44,9 @@ pipeline {
             }
             steps {
               withCredentials([string(credentialsId: "metal3-clusterctl-github-token", variable: "GITHUB_TOKEN")]) {
-                sh "./hack/ci-e2e.sh"
+                timestamps {
+                  sh "./hack/ci-e2e.sh"
+                }
               }
             }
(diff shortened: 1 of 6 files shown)


jenkins/jobs/bmo_e2e_tests.pipeline
```diff
@@ -56,9 +58,11 @@ pipeline {
                       currentBuild.result = "FAILURE"
                   }
                 }
-                archiveArtifacts "artifacts*.tar.gz"
-                /* Clean up */
-                sh "make clean-e2e"
+                timestamps {
+                  archiveArtifacts "artifacts*.tar.gz"
```
tuminoid, line 62: Do we want archiveArtifacts to be in here? Elsewhere its outside timestamps.

Sunnatillo (commented):
I checked the logs. This is amazing. Thank you @Rozzii. 
LGTM

jenkins/jobs/bmo_e2e_tests.pipeline
Rozzii: no not really I think we are fine without it

tuminoid (commented):
/approve

Thanks, this will help a lot in debugging.

PR #615, opened 2024-01-09
Make BMO e2e tests required status check

lentzi90:
I propose we make the e2e tests required for BMO now. They exist on main and release-0.5 so this is what I added.
We could also remove one or both CAPM3 based jobs, but perhaps better to do that separately?


lentzi90:
Irrelevant...

maxrantil:
LGTM

PR #618, opened 2024-01-17
Trigger BMO e2e tests through prow

lentzi90:
This is to verify that the jenkins operator for prow works as expected. The idea is to trigger jenkins jobs through prow instead of through the ghprb jenkins plugin. When we have verified that this works, we can continue with the rest of the jenkins jobs and remove the ghprb triggers.

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -540,6 +540,16 @@ presubmits:
                 value: "/"
             image: ghcr.io/yannh/kubeconform:v0.6.2-alpine@sha256:49b5f6b320d30c1b8b72a7abdf02740ac9dc36a3ba23b934d1c02f7b37e6e740
             imagePullPolicy: Always
+    - name: metal3-bmo-e2e-test-pull
+      # E2e tests do not exist before release-0.5
+      branches:
+        - main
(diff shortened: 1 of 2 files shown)


lentzi90:
/hold
This needs https://gerrit.nordix.org/c/infra/cicd/+/20350

prow/manifests/overlays/metal3/config.yaml
```diff
@@ -138,11 +138,7 @@ branch-protection:
                   [
                     "test-centos-e2e-integration-main",
                     "test-ubuntu-integration-main",
-                    "metal3-bmo-e2e-test",
                   ]
-            release-0.5:
-              required_status_checks:
-                contexts: ["metal3-bmo-e2e-test"]
```
lentzi90, line 145: Not sure if we should really remove these. I marked the job as not optional below but I'm not sure if that means it will show a status context. If it doesn't show the context we should add these back.

lentzi90:
Changed it now to keep the status checks and make the prow job optional. Once we have verified it is working I can update these things.

lentzi90:
Irrelevant

tuminoid (commented):
/approve

Sunnatillo (commented):
/lgtm

PR #621, opened 2024-01-18
Make BMO e2e prow job required

lentzi90:
This removes the required status context for BMO e2e tests and instead makes the prow job required.

It also removes the report template because this is already covered by the report template set for plank. This config is currently causing double messages when a job fails (one for jenkins and one for plank).

One thing to discuss, test and decide is how to handle these required jobs.
Other required jobs that we have run automatically when needed. The e2e jobs have traditionally been manually triggered only to save some resources. I'm not sure if a required but not automatically run job will show up in the status contexts. If not, we have to decide if we

1. Add required status contexts back for them, or
2. Run the jobs automatically. If we choose this we can also use config like "skip if only" to not run it on all PRs, which would be quite nice!

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -138,11 +138,7 @@ branch-protection:
                   [
                     "test-centos-e2e-integration-main",
                     "test-ubuntu-integration-main",
-                    "metal3-bmo-e2e-test",
                   ]
-            release-0.5:
-              required_status_checks:
-                contexts: ["metal3-bmo-e2e-test"]
         cluster-api-provider-metal3:
           branches:
             main:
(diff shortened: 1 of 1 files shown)


prow/manifests/overlays/metal3/config.yaml
```diff
@@ -542,7 +533,15 @@ presubmits:
       agent: jenkins
       # Don't run unless triggered to avoid wasting resources
       always_run: false
-      # Until we have checked that it works, keep it optional
+      optional: false
```
lentzi90, line 536: What do you think? Should we trigger it automatically instead but have 
```
skip_if_only_changed: '(((^|/)OWNERS)|(\.md))$'
```

tuminoid: I think this is the way to go. We've spent quite a lot of time developing the e2e test to make BMO testing more standalone, and its rather fast and stable test, so why wouldn't we want to run it?

Rozzii (approved):
/approve

lentzi90:
/hold
@Rozzii what do you think about changing it to use `skip_if_only_changed`?

tuminoid (commented):
/approve

Rozzii:
> /hold @Rozzii what do you think about changing it to use `skip_if_only_changed`?

If the regexp works and it correctly skips the md and the OWNER files then I am fine with it.

lentzi90:
After offline discussion, I think we can keep the old behavior of manual trigger but required job.
After merging this I will check if we need the required status check or not and do a follow up PR if needed.
/hold cancel

PR #631, opened 2024-02-01
Prow: Bump ingress-nginx to v1.9.6

lentzi90:
Just keeping up with the upstream. Nothing newsworthy in the changelog.


tuminoid (commented):
/approve

Already applied in the cluster right?

lentzi90:
> Already applied in the cluster right?

It is applied now! :slightly_smiling_face:

PR #641, opened 2024-02-27
Prow: Restore patches that were mistakenly removed

lentzi90:
The patches used to add resource requests for all prow containers were mistakenly removed when adding the autobumper. This commits restores them.

(The files that they refer to are still there.)


kashifest (commented):
/approve
/lgtm

PR #642, opened 2024-02-28
Add new pipeline to test triggering basic e2e from prow

mboukhalfa:


adilGhaffarDev:
we are adding a new pipeline script anyhow, can we use dynamic workers here while we are at it?

mboukhalfa:
> dding a new pipeline script anyhow, can we use dynamic workers here while we are at it?
This is just testing the changes with basic e2e once we get this working we can re draw a design in term of the dynamic workers and the pipeline separation

jenkins/jobs/prow_integration_tests.pipeline
Sunnatillo, line 1: Why the name is prow integration tests? I suggest renaming it basic_e2e_integration_tests.pipeline

Sunnatillo (commented):
LGTM, one nit

jenkins/jobs/prow_integration_tests.pipeline
mboukhalfa: it is not specific for basic now I just used the same one existing `integration_tests.test ` with changes needed to be triggered from prow

mboukhalfa: What do you suggest ?

Sunnatillo: We had a discussion on a call. This will be changed later, for now we test it first

PR #645, opened 2024-03-01
Check CI VM cloud-init exit code instead of output

lentzi90:
The output can contain text that matches even though no fatal error occurred. Since we no longer have the issue where centos is always failing cloud-init, we can now actually check the exit code.

### DIFF
--- a/jenkins/scripts/utils.sh
+++ b/jenkins/scripts/utils.sh
@@ -60,15 +60,11 @@ vm_healthy()
     KEY="${2:?}"
     SERVER="${3:?}"
 
-    cloud_init_status=$(ssh -o ConnectTimeout=2 -o StrictHostKeyChecking=no \
+    if ssh -o ConnectTimeout=2 -o StrictHostKeyChecking=no \
         -o UserKnownHostsFile=/dev/null -i "${KEY}" \
-        "${USER}"@"${SERVER}" cloud-init status --long --wait)
-    if echo "${cloud_init_status}" | grep "error"; then
-        echo "There was a cloud-init error:"
-        echo "${cloud_init_status}"
-        return 1
-    else
-        echo "Cloud-init completed successfully!"
+        "${USER}"@"${SERVER}" cloud-init status --long --wait; then
         return 0
+    else
+        return 1
     fi
 }
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -127,11 +127,10 @@ fi
 echo "Waiting for the host ${TEST_EXECUTER_VM_NAME} to come up"
 # Wait for the host to come up
 wait_for_ssh "${METAL3_CI_USER}" "${METAL3_CI_USER_KEY}" "${TEST_EXECUTER_IP}"
-# FIXME (lentzi90): Temporarily disabled to get past warnings in cloud-init
-# if ! vm_healthy "${METAL3_CI_USER}" "${METAL3_CI_USER_KEY}" "${TEST_EXECUTER_IP}"; then
-#   echo "Server is unhealthy. Giving up."
-#   exit 1
-# fi
+if ! vm_healthy "${METAL3_CI_USER}" "${METAL3_CI_USER_KEY}" "${TEST_EXECUTER_IP}"; then
+  echo "Server is unhealthy. Giving up."
+  exit 1
+fi
 
 TEMP_FILE_NAME=$(mktemp vars-XXXXXX.sh)
 cat <<-EOF >>"${CI_DIR}/files/${TEMP_FILE_NAME}"


jenkins/scripts/utils.sh
```diff
@@ -60,15 +60,11 @@ vm_healthy()
...
     SERVER="${3:?}"
 
-    cloud_init_status=$(ssh -o ConnectTimeout=2 -o StrictHostKeyChecking=no \
+    if ssh -o ConnectTimeout=2 -o StrictHostKeyChecking=no \
         -o UserKnownHostsFile=/dev/null -i "${KEY}" \
-        "${USER}"@"${SERVER}" cloud-init status --long --wait)
-    if echo "${cloud_init_status}" | grep "error"; then
-        echo "There was a cloud-init error:"
-        echo "${cloud_init_status}"
-        return 1
```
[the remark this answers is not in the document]
tuminoid: Not every word "error" means fatal. Instead, if the cloud-init exits with error, it is considered fatal.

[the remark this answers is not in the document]
lentzi90: No I just simplified it so that we run it directly in the `if`. When it returns with exit code 0 that counts as `if true`, when it returns something else it counts as `if false`.
It will also output directly everything so we can see it when it runs.

jenkins/scripts/utils.sh
```diff
@@ -60,15 +60,11 @@ vm_healthy()
     KEY="${2:?}"
     SERVER="${3:?}"
 
-    cloud_init_status=$(ssh -o ConnectTimeout=2 -o StrictHostKeyChecking=no \
+    if ssh -o ConnectTimeout=2 -o StrictHostKeyChecking=no \
```
tuminoid, line 63: Noticed that `cloud-init` might return 1 for fatal, or 2 for recoverable error. This converts both to fatals. Just making sure this is what we want?

jenkins/scripts/utils.sh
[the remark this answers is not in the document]
kashifest: you mean even if cloud_init_status has errors word, checking the status code itself is enough?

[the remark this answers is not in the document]
tuminoid: All this applies of course only if the cloud-init script contains `set -e` or is run in a way that it can fail. Otherwise it will just return last commands return value, which would be misleading.

jenkins/scripts/utils.sh
tuminoid:
```console
19:35:58  SSH connection to host[10.100.10.180] is up.
19:35:58  Warning: Permanently added '10.100.10.180' (ED25519) to the list of known hosts.
19:36:05  errors: []
19:36:05  recoverable_errors:
19:36:05  There was a cloud-init error:
19:36:05  ...........................
19:36:05  status: done
19:36:05  extended_status: degraded done
19:36:05  boot_status_code: enabled-by-generator
19:36:05  last_update: Thu, 29 Feb 2024 17:36:04 +0000
19:36:05  detail:
19:36:05  DataSourceOpenStackLocal [net,ver=2]
19:36:05  errors: []
19:36:05  recoverable_errors:
19:36:05  WARNING:
19:36:05  	- Invalid network-config provided: config.0.mtu: None is not of type 'integer' config.1.address: '1.1.1.1' is not of type 'array' config.2.address: '8.8.8.8' is not of type 'array'
19:36:05  Server is unhealthy. Giving up.
```
This would actually be the exact case here. The error is recoverable warning, so the return value would be 2, but this check would treat it as fatal.

kashifest: Good catch @tuminoid

lentzi90: Oh I had missed that actually! I think we are fine with it this way. Checking the logs it looks like it didn't count the warning as a recoverable error:
```
10:44:57  status: done
10:44:57  extended_status: degraded done
10:44:57  boot_status_code: enabled-by-generator
10:44:57  last_update: Fri, 01 Mar 2024 08:44:57 +0000
10:44:57  detail:
10:44:57  DataSourceOpenStackLocal [net,ver=2]
10:44:57  errors: []
10:44:57  recoverable_errors:
10:44:57  WARNING:
10:44:57  	- Invalid network-config provided: config.0.mtu: None is not of type 'integer' config.1.address: '1.1.1.1' is not of type 'array' config.2.address: '8.8.8.8' is not of type 'array'
```

I confirmed also by sshing to the VM:

```
metal3ci@ci-test-vm-20240301084413-hxzq:~$ cloud-init status --long 
status: done
extended_status: degraded done
boot_status_code: enabled-by-generator
last_update: Fri, 01 Mar 2024 08:44:57 +0000
detail:
DataSourceOpenStackLocal [net,ver=2]
errors: []
recoverable_errors:
WARNING:
	- Invalid network-config provided: config.0.mtu: None is not of type 'integer' config.1.address: '1.1.1.1' is not of type 'array' config.2.address: '8.8.8.8' is not of type 'array'
metal3ci@ci-test-vm-20240301084413-hxzq:~$ echo $?
0

```

lentzi90: I guess we should report to Cleura also that we get invalid network-config...

tuminoid: It seems Ubuntu 22.04 does not have error 2 anymore. It either fails or succeeds, and recoverable error is success.
However, Ubuntu 20.04 and Centos do have error 2 still, which this will turn into fatal.

tuminoid:
@lentzi90 are you going to change the implementation or shall we proceed as is?

lentzi90:
Oh I had forgotten about this.
I think it is good like this. Any objections to merge as is?

kashifest (commented):
/approve

tuminoid (commented):
/lgtm

PR #646, opened 2024-03-01
Run dev-env's 01 scripts in all integration tests

kashifest:
Skipping 01 scripts has caused quite a few issues lately. Lets run it in all ansible integration tests to avoid confusion.


tuminoid (commented):
While it is mostly redundant, the confusion is worse and it allows changes to go thru untested / makes them untestable.

/approve

lentzi90 (commented):
/lgtm

PR #650, opened 2024-03-01
Fix UPDATED_REPO in the pipeline for e2e test trigger from prow

mboukhalfa:
This supposed to be the full link but in the environment vlock it was initiated to only repo name this PR fix that and reorder the vars


tuminoid (commented):
/approve

lentzi90 (commented):
/lgtm

PR #660, opened 2024-03-05
fix BMO required tests to match GH settings

tuminoid:
Fix BMO required tests to match reality.

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -134,7 +134,21 @@ branch-protection:
               required_status_checks:
                 contexts:
                   [
-                    "test-centos-e2e-integration-main",
+                    "test-centos-e2e-integration-main"
+                  ]
+            release-0.5:
+              required_status_checks:
+                contexts:
+                  [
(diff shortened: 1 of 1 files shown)


smoshiur1237 (commented):
/lgtm

prow/manifests/overlays/metal3/config.yaml
```diff
@@ -135,6 +135,22 @@ branch-protection:
                 contexts:
                   [
                     "test-centos-e2e-integration-main",
+                    "metal3-bmo-e2e-test-pull"
```
lentzi90, line 138: I think the linter is complaining about this since it is a prow job. There is no need to have it in the required contexts when it is set as a postsubmit further down

lentzi90 (commented):
/approve

prow/manifests/overlays/metal3/config.yaml
tuminoid: Yeah, read the comment that this list is only for prow-external tests, which bmo e2e is not. Removed.

smoshiur1237 (commented):
/lgtm

PR #662, opened 2024-03-05
Adds number on install script's name

Sunnatillo:
devuser elements script has 50 in the name, our install scripts should run after user has been created. 
This fixes the issue of `user is not added to docker group` because it was not created. Current change makes sure our install script runs after devuser element.


lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #664, opened 2024-03-07
Set jenkins worker for BML job

Sunnatillo:
This PR:
- sets new jenkins worker (vm inside a Jumphost) to be used for bml jobs 
- removes Jumphost related logic from bml scripts
- changes jenkins agent label for bml jobs


tuminoid (commented):
/lgtm

PR #675, opened 2024-03-20
Add feature tests triggers to prow config

mboukhalfa:
Part of migrating tests to be triggerable from prow using prow jenkins operator.This PR updates prow config with triggers


kashifest (commented):
/lgtm

PR #676, opened 2024-03-22
Use dhcp provided dns address for metal3ci ubuntu images

Sunnatillo:
in centos images we use dns servers that are provided by openstack network, This PR makes ubuntu ci images to use dns provided dns names


tuminoid (commented):
/approve

PR #677, opened 2024-03-25
Fetch changes from PR id with prow feature tests triggers

mboukhalfa:
Prow jenkins operator does not propagate vars to the sources branch of the PR so we have to pull changes form the PR id

### DIFF
--- a/jenkins/scripts/dynamic_worker_workflow/feature_tests.sh
+++ b/jenkins/scripts/dynamic_worker_workflow/feature_tests.sh
@@ -68,6 +68,10 @@ if [[ "${UPDATED_REPO}" != *"${REPO_ORG}/${REPO_NAME}"* ]] ||
     git config user.name "Test"
     git remote add test "${UPDATED_REPO}"
     git fetch test
+    # if triggered from prow we cannot get the ghprbAuthorRepoGitUrl then we pull the PR
+    if [[ -n "${PR_ID:-}" ]]; then
+        git fetch origin "pull/${PR_ID}/head:${UPDATED_BRANCH}-branch" || true
+    fi
     # Merging the PR with the target branch
     git merge "${UPDATED_BRANCH}" || exit
 fi
--- a/jenkins/scripts/files/run_integration_tests.sh
+++ b/jenkins/scripts/files/run_integration_tests.sh
@@ -123,7 +123,9 @@ if [[ "${UPDATED_REPO}" != *"${REPO_ORG}/${REPO_NAME}"* ]] ||
     git remote add test "${UPDATED_REPO}"
     git fetch test
     # if triggered from prow we cannot get the ghprbAuthorRepoGitUrl then we pull the PR
-    git fetch origin "pull/${PR_ID:-0}/head:${UPDATED_BRANCH}-branch" || true
+    if [[ -n "${PR_ID:-}" ]]; then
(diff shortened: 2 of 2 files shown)


smoshiur1237 (commented):
/lgtm

jenkins/scripts/dynamic_worker_workflow/feature_tests.sh
```diff
@@ -68,6 +68,8 @@ if [[ "${UPDATED_REPO}" != *"${REPO_ORG}/${REPO_NAME}"* ]] ||
     git config user.name "Test"
     git remote add test "${UPDATED_REPO}"
     git fetch test
+    # if triggered from prow we cannot get the ghprbAuthorRepoGitUrl then we pull the PR
+    git fetch origin "pull/${PR_ID:-0}/head:${UPDATED_BRANCH}-branch" || true
```
tuminoid, line 72: I don't like the idea of doing invalid git fetches, can we put this in conditional so we'd only pull when we have `$PR_ID`?

mboukhalfa: Here we are defaulting the var to 0 if null then should we just check if the PR_ID is not zero to run this ?

tuminoid: PR number 0 makes it invalid to GH. We're calling external service with invalid ID, hence skipping it if PR_ID is 0 would be preferable.

jenkins/scripts/dynamic_worker_workflow/feature_tests.sh
```diff
@@ -68,6 +68,10 @@ if [[ "${UPDATED_REPO}" != *"${REPO_ORG}/${REPO_NAME}"* ]] ||
     git config user.name "Test"
     git remote add test "${UPDATED_REPO}"
     git fetch test
+    # if triggered from prow we cannot get the ghprbAuthorRepoGitUrl then we pull the PR
+    if [[ -n "$PR_ID" ]]; then
```
tuminoid, line 72:
```suggestion
    if [[ -n "${PR_ID:-}" ]]; then
```
Otherwise it'll fail to `set -u` being unset.

jenkins/scripts/files/run_integration_tests.sh
```diff
@@ -123,7 +123,9 @@ if [[ "${UPDATED_REPO}" != *"${REPO_ORG}/${REPO_NAME}"* ]] ||
     git remote add test "${UPDATED_REPO}"
     git fetch test
     # if triggered from prow we cannot get the ghprbAuthorRepoGitUrl then we pull the PR
-    git fetch origin "pull/${PR_ID:-0}/head:${UPDATED_BRANCH}-branch" || true
+    if [[ -n "${PR_ID}" ]]; then
```
tuminoid, line 126:
```suggestion
    if [[ -n "${PR_ID:-}" ]]; then
```
Same

mboukhalfa: done

jenkins/scripts/dynamic_worker_workflow/feature_tests.sh
mboukhalfa: done

tuminoid (commented):
/approve

smoshiur1237 (commented):
/lgtm

PR #684, opened 2024-04-03
Prow: Add infra node pool

lentzi90:
This adds a separate node pool for everything that isn't a prow job or
running on the control-plane. The node pool has a taint to prevent
workload running there by accident. All relevant workload now has
toleration for this taint and a node selector to always run in this node
pool.


lentzi90:
This has been successfully applied in the cluster.

/hold cancel

tuminoid (commented):
/approve

PR #685, opened 2024-04-03
Fix upgrade test pipeline and set the trigger branches

mboukhalfa:
Specify branches on capm3 and bmo where tests should run and fix typo in the pipeline


tuminoid (commented):
/approve

PR #687, opened 2024-04-04
Update BMO e2e matrix

lentzi90:
We now have support for redfish-virtualmedia i BMO e2e tests. The tests as ow now, check redfish-virtualmedia and ipmi in parallel. This commit adds redfish to the matrix and configures it so that periodic jobs would check all three protocols and PRs stay as they are.


lentzi90:
/hold
I need to check that the syntax is OK

lentzi90:
Confirmed working here: https://jenkins.nordix.org/blue/organizations/jenkins/metal3-bmo-e2e-test-pull/detail/metal3-bmo-e2e-test-pull/382/pipeline/
/hold cancel

I expect to simplify this still by changing the parameters so that we can have a single one just for the BMC protocol. Then we can in BMO pick VBMC/sushy-tools based on that. This PR is just a step on the way to do that without breaking things.

lentzi90:
For now the `SUSHY_TOOLS_PROTOCOL` variable is just overwritten in BMO when VBMC is used. So please disregard the `BMO_E2E_EMULATOR=vbmc` + `SUSHY_TOOLS_PROTOCOL=redfish-virtualmedia`. It is just nonsense. We must specify both parameters so that is why it is there.

smoshiur1237 (commented):
/lgtm

jenkins/jobs/bmo_e2e_tests.pipeline
```diff
@@ -19,14 +19,38 @@ pipeline {
...
+        }
+        excludes {
+          exclude {
+            axis {
+              name 'BMO_E2E_EMULATOR'
+              values 'vbmc'
+            }
+            axis {
+              name 'SUSHY_TOOLS_PROTOCOL'
+              values 'redfish', 'redfish-multimedia'
```
tuminoid, line 47: Is this correct? `redfish-multimedia` is not mentioned elsewhere in Metal3, and even google doesn't give results for that.

lentzi90: Oh my. Thanks for spotting that!
I was expecting it not to work when I filtered out both redfish and redfish-virtualmedia. Got surprised when it worked but didn't look closer. This looks like the explanation :sweat_smile:

tuminoid (commented):
/lgtm

kashifest (commented):
/approve

PR #691, opened 2024-04-09
Simplify BMO e2e pipeline matrix

lentzi90:
We can now set the BMC protocol directly instead of specifying the emulator and optionally the sushy-tools protocol.


kashifest (commented):
/lgtm

PR #694, opened 2024-04-10
Build ironic locally when running tests on ironic-image repo

mboukhalfa:
There is a behavior implimented in old ansible tests to use postfix that make the test building ironic locally even when running on a different repo but it does not seem being used

### DIFF
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -54,13 +54,10 @@ KUBERNETES_VERSION_UPGRADE_FROM="${KUBERNETES_VERSION_UPGRADE_FROM:-}"
 KUBERNETES_VERSION_UPGRADE_TO="${KUBERNETES_VERSION_UPGRADE_TO:-}"
 KUBECTL_SHA256="${KUBECTL_SHA256:-}"
 
-if [[ "${IRONIC_INSTALL_TYPE}" == "source" ]]; then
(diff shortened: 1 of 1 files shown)


jenkins/scripts/integration_test.sh
```diff
@@ -54,13 +54,10 @@ KUBERNETES_VERSION_UPGRADE_FROM="${KUBERNETES_VERSION_UPGRADE_FROM:-}"
 KUBERNETES_VERSION_UPGRADE_TO="${KUBERNETES_VERSION_UPGRADE_TO:-}"
 KUBECTL_SHA256="${KUBECTL_SHA256:-}"
 
-if [[ "${IRONIC_INSTALL_TYPE}" == "source" ]]; then
+# Build ironic locally when running tests on ironic-image repo  
+if [[ "${REPO_NAME}" == "ironic-image" ]]; then
```
tuminoid, line 58: We do not want to always build from source. RPM installation is the default method, source building is special case.

We need to have source build triggerable via a keyword for those special cases, and default should stay RPM based install. It will still build the ironic container with the changes via `tested_repo`, as we can see from the e2e keyword tested yesterday, since https://github.com/metal3-io/ironic-image/pull/495 failed as it depends on https://github.com/metal3-io/ironic-image/pull/493 which succeeded.

/hold

mboukhalfa:
tests are working as expected building ironic from main by default will be considered

PR #696, opened 2024-04-12
update ironic-image required tests to Prow tests

tuminoid:
Update ironic-image required tests to Prow tests. Remove optional tests as they conflict with required tests, and make little sense to be available cross-releases.

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -169,36 +169,6 @@ branch-protection:
             release-1.6:
               required_status_checks:
                 contexts: ["test-ubuntu-e2e-integration-release-1-6"]
-        ironic-image:
-          branches:
-            main:
-              required_status_checks:
-                contexts:
-                  [
-                    "test-ubuntu-integration-main",
-                    "test-centos-integration-main",
-                  ]
-            release-23.1:
-              required_status_checks:
-                contexts:
-                  [
-                    "test-ubuntu-integration-main",
-                    "test-centos-integration-main",
-                  ]
-            release-24.0:
-              required_status_checks:
-                contexts:
-                  [
-                    "test-ubuntu-integration-main",
-                    "test-centos-integration-main",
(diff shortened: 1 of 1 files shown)


tuminoid:
/hold

Optional tests needed to be removed as they conflict with required tests. IMO they also make little sense anyways. Thoughts?

/cc @mboukhalfa @kashifest @lentzi90

prow/manifests/overlays/metal3/config.yaml
```diff
@@ -2921,39 +2921,6 @@ presubmits:
                 value: "TRUE"
             image: docker.io/pipelinecomponents/markdownlint:0.13.0@sha256:9c0cdfb64fd3f1d3bdc5181629b39c2e43b6a52fc9fdc146611e1860845bbae0
             imagePullPolicy: Always
-    # name: {job_prefix}-{image_os}-e2e-integration-test-{capm3_target_branch}
-    - name: metal3-centos-e2e-integration-test-main
-      agent: jenkins
-      always_run: false
-      optional: true
```
lentzi90, line 2928: These are the actual prow job definitions. Without them you won't be able to trigger anything through prow.
The required status checks above are just things tide should make sure are there before merging. I.e. prow knows nothing about them except that they must be there

lentzi90 (changes requested):
The prow jobs have to stay if we want to trigger the tests through prow.

prow/manifests/overlays/metal3/config.yaml
tuminoid: Oh yes, of course. However they cannot be optional, and need to be branch by branch defined then, as they cannot be listed like this and be required. Prow won't allow that.

tuminoid:
OK, fixed the confusion with the jenkins contexts and prow contexts. How about now?

lentzi90 (commented):
/approve

Rozzii (approved):
/lgtm

PR #700, opened 2024-04-15
Make two tests required on main and only ubuntu on releases

mboukhalfa:
Update prow required tests


kashifest (commented):
/approve

smoshiur1237 (commented):
/lgtm

PR #701, opened 2024-04-16
Make e2e run on other repos from prow

mboukhalfa:
Add prow triggered test on the following repos
ironic-ipa-downloader
mariadb-image


kashifest (commented):
/approve

PR #706, opened 2024-04-18
WIP: add branchprotector cronjob to prow

tuminoid:
Adding branchprotector cronjob to Prow to automatically handle branch protection settings for all repos instead of manual configuration.

https://docs.prow.k8s.io/docs/components/optional/branchprotector/#deploy-cronjob-to-production


tuminoid:
For now, due shortcomings in branchprotector, we'll not go for it.

/close

PR #709, opened 2024-04-23
Fail on pipe failure for openstack commands

kashifest:
as the title says

### DIFF
--- a/jenkins/scripts/integration_test.sh
+++ b/jenkins/scripts/integration_test.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -eu
+set -o pipefail
 
 # Description:
 #   Runs the integration tests for metal3-dev-env in an executer vm


smoshiur1237 (commented):
/lgtm

mquhuy:
With this the script will fail as soon as something fails. If there were resources created before that, those resources wouldn't be cleaned.
Should we add a cleanup trap?

tuminoid:
> With this the script will fail as soon as something fails. If there were resources created before that, those resources wouldn't be cleaned. Should we add a cleanup trap?

1. `set -e` has been there always, just adding `-o pipefail` does not change it. It may just trigger in few more commands.
2. Cleanup is AFAIK done in the pipeline, and trap is not needed.

kashifest:
> With this the script will fail as soon as something fails. If there were resources created before that, those resources wouldn't be cleaned. Should we add a cleanup trap?

Don't we have that already in the pipeline https://github.com/metal3-io/project-infra/blob/main/jenkins/jobs/integration_tests.pipeline#L149

tuminoid (commented):
/approve

PR #716, opened 2024-04-26
Add libvirt related packages for CI image

Sunnatillo:
Add left out packages related to libvirt when we move to DiskImage builder

### DIFF
--- a/jenkins/image_building/dib_elements/ci-base/pkg-map
+++ b/jenkins/image_building/dib_elements/ci-base/pkg-map
@@ -9,10 +9,12 @@
     "family": {
         "redhat": {
           "bash-completion":"bash-completion",
+          "build-essential":"",
           "curl": "curl",
           "dnsmasq":"dnsmasq",
           "git": "git",
           "libguestfs-tools":"libguestfs-tools",
+          "libvirt-daemon-system":"",
           "make": "make",
           "openjdk-11-jre": "java-11-openjdk",
           "ovmf":"edk2-ovmf",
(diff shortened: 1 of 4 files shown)


lentzi90 (commented):
/approve

jenkins/image_building/dib_elements/ci-base/pkg-map
```diff
@@ -21,14 +23,17 @@
...
           "wget": "wget"
         },
         "debian":{
           "bash-completion":"bash-completion",
+          "build-essential":"build-essential",
           "curl": "curl",
           "dnsmasq":"dnsmasq",
           "git": "git",
           "libguestfs-tools":"libguestfs-tools",
+          "libvirt-daemon-system":"",
```
tuminoid, line 36: Why we're adding this in the first place, since on both redhat and debian it is mapped to `""` ?

tuminoid (commented):
Question

tuminoid:
Also, amend the title please. You're not modifying Ubuntu only.
It would also nice to see proper description why the packages are needed.

jenkins/image_building/dib_elements/ci-base/pkg-map
lentzi90: Oh I missed that. Why is it mapper to ""?

Sunnatillo: fixed

tuminoid:
Do we have tests for testing this?

/test ?

lentzi90:
> @lentzi90 why do those packages needed? are they specifically for BMO tests?

They are needed for the BMO e2e tests. I would also say that most of them are also needed for other tests, it's just that dev-env installs them automatically.
Note that this is not something new. These packages were installed by the old packer-based scripts. I don't know why they were forgotten when moving to DIB.

smoshiur1237 (approved):
/lgtm

PR #719, opened 2024-04-29
Prow: Bump infra and cluster-resources

lentzi90:
This bumps the version of
- calico to v3.27.3
- cloud-provider-openstack and the CSI plugin to v1.29.0
- ingress-nginx to v1.10.1
- CAPI to v1.7.1
- CAPO to v0.10.0

Note that we do not have declarative config for CAPI and CAPO so the changes for them are just for the documentation.

Changes are already applied.

### DIFF
--- a/prow/README.md
+++ b/prow/README.md
@@ -341,7 +341,8 @@ images (see sections above).
 
    ```bash
    kind create cluster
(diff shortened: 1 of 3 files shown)


kashifest (commented):
/approve

smoshiur1237 (commented):
/lgtm

PR #720, opened 2024-04-29
Prow: Update CAPO resources to use v1beta1 API

lentzi90:
The latest release of CAPO came with a new API version. This commit updates all manifests to use that version. This is essentially a no-op in the cluster as all resources are anyway converted back and forth automatically by the webhook as needed.

The biggest change are the managed security groups. They used to be hard-coded in CAPO to match what is needed for Calico. Now they must be specified explicitly.

### DIFF
--- a/prow/capo-cluster/cluster.yaml
+++ b/prow/capo-cluster/cluster.yaml
@@ -13,6 +13,6 @@ spec:
     kind: KubeadmControlPlane
     name: prow-control-plane
   infrastructureRef:
-    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha6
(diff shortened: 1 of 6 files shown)


kashifest:
> This is essentially a no-op in the cluster as all resources are anyway converted back and forth automatically by the webhook as needed.

Meaning ? did we do something similar to `clusterctl upgrade ...` or we didnt here ?

lentzi90:
Not clusterctl upgrade. I did `kubectl diff` to see that it caused no changes* and then `kubectl apply`.

*) In fact it changes the description of the security groups from `Created by cluster-api-provider-openstack API conversion - <specific-rule>` to `Calico <specific-rule>`.

Sunnatillo (commented):
/lgtm

PR #731, opened 2024-05-03
🌱 Fix dev-env triggers

mboukhalfa:
Add prefix to the triggers and add the missing triggers to the some repos and remove them where not needed


Sunnatillo (commented):
/lgtm

lentzi90 (commented):
/approve

PR #732, opened 2024-05-03
Unset null and empty vars that are coming from jjb

Sunnatillo:
Apparently if variable does not come from jjb, pipeline sets it to null value. This null value is not detected by  `-z "${VAR:-}"` condition. 

This PR changes conditions to `-z ${"VAR" == "null"}` to detect unset variable

### DIFF
--- a/jenkins/scripts/dynamic_worker_workflow/e2e_tests.sh
+++ b/jenkins/scripts/dynamic_worker_workflow/e2e_tests.sh
@@ -39,24 +39,33 @@ if [[ "${CAPM3_VERSION}" == "v1alpha5" ]]; then
     export KUBERNETES_VERSION="v1.23.8"
     export KUBECTL_SHA256="${KUBECTL_SHA256:-4685bfcf732260f72fce58379e812e091557ef1dfc1bc8084226c7891dd6028f}"
 fi
-if  [[ -z "${KUBERNETES_VERSION:-}" ]]; then
+
+if [[ -z "${NUM_NODES:-}" ]] || [[ "${NUM_NODES}" == "null" ]]; then
+    unset NUM_NODES
+fi
+
+if [[ -z "${GINKGO_FOCUS:-}" ]] || [[ "${GINKGO_FOCUS}" == "null" ]]; then
+    unset GINKGO_FOCUS
+fi
+
+if [[ -z "${GINKGO_SKIP:-}" ]] || [[ "${GINKGO_SKIP}" == "null" ]]; then
     unset GINKGO_SKIP
 fi
 
-if  [[ -z "${KUBECTL_SHA256:-}" ]]; then
-    unset KUBECTL_SHA256
+if [[ -z "${EPHEMERAL_TEST:-}" ]] || [[ "${EPHEMERAL_TEST}" == "null" ]]; then
+    unset EPHEMERAL_TEST
 fi
 
-if  [[ -z "${KUBERNETES_VERSION_UPGRADE_FROM:-}" ]]; then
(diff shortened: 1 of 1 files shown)


Sunnatillo:
manual test:
https://jenkins.nordix.org/job/metal3-e2e-clusterctl-upgrade-test-main/29/console

jenkins/scripts/dynamic_worker_workflow/e2e_tests.sh
```diff
@@ -39,24 +39,33 @@ if [[ "${CAPM3_VERSION}" == "v1alpha5" ]]; then
...
 fi
 
-if  [[ -z "${KUBERNETES_VERSION_UPGRADE_TO:-}" ]]; then
+if [[ "${KUBERNETES_VERSION_UPGRADE_TO}" == "null" ]]; then
     unset KUBERNETES_VERSION_UPGRADE_TO
 fi
 
-if  [[ -z "${EPHEMERAL_TEST:-}" ]]; then
-    unset EPHEMERAL_TEST
+if [[ "${KUBECTL_SHA256}" == "null" ]]; then
```
NymanRobin, line 67: An idea could be to configure all environment variables always as a parameter in Jenkins job then we could have the default as empty string or null. In my opinion that would be a bit more explicit and work a bit as a documentation as currently it was somewhat confusing at least to me from where this null came. What do you think about that?

NymanRobin (commented):
Maybe there should be a standard for this or some kind of documentation as I see this can easily happen again otherwise?

jenkins/scripts/dynamic_worker_workflow/e2e_tests.sh
Sunnatillo: That is the issue of null value in groove and in shell scripts. When value is not defined in pipeline script it is set to null, and that null is converted as a string null into shell. I am not sure if we can fix that on jenkins side

jenkins/scripts/dynamic_worker_workflow/e2e_tests.sh
```diff
@@ -39,24 +39,33 @@ if [[ "${CAPM3_VERSION}" == "v1alpha5" ]]; then
     export KUBERNETES_VERSION="v1.23.8"
     export KUBECTL_SHA256="${KUBECTL_SHA256:-4685bfcf732260f72fce58379e812e091557ef1dfc1bc8084226c7891dd6028f}"
 fi
-if  [[ -z "${KUBERNETES_VERSION:-}" ]]; then
+
+if [[ "${NUM_NODES}" == "null" ]]; then
```
tuminoid, line 43:
```suggestion
if [[ -z "${NUM_NODES:-}" ]] || [[ "${NUM_NODES}" == "null" ]]; then
```
Better to have these all also not explode on undefined values.

tuminoid (commented):
IMO its better this way so it doesn't explode on unset/empty value.

tuminoid:
We also have this issue break this PR: https://github.com/metal3-io/baremetal-operator/pull/1718

jenkins/scripts/dynamic_worker_workflow/e2e_tests.sh
NymanRobin: I meant defining them in Jenkins parameters and I think that should be able to fix it with a default value ("" or null) there see below, but actually I like @tuminoid idea more it is simpler. Maybe make it a function and the function can document why it is done?

image

jenkins/scripts/dynamic_worker_workflow/e2e_tests.sh
Sunnatillo: agreed

tuminoid (commented):
/lgtm

smoshiur1237 (approved):
LGTM

mquhuy:
@Sunnatillo Could you please update the PR description with the current implementation?

PR #737, opened 2024-05-07
WIP: add affected branch to all tests

tuminoid:
This is try limit branchprotector from wandering off to feature branches etc. If all tests have branches, it shouldn't go and try apply the settings to ALL branches it can find.

WIP:
- need to add settings to match our desired state to the "branch-protector:" section, like disable admin bypass, require PR etc. Branchprotector will set all those as well.

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -226,6 +226,8 @@ periodics:
 presubmits:
   Nordix/metal3-clusterapi-docs:
   - name: shellcheck
+    branches:
+    - main
     run_if_changed: '((\.sh)|^Makefile)$'
     decorate: true
     spec:
@@ -240,6 +242,8 @@ presubmits:
         image: docker.io/koalaman/shellcheck-alpine:v0.10.0@sha256:5921d946dac740cbeec2fb1c898747b6105e585130cc7f0602eec9a10f7ddb63
         imagePullPolicy: Always
   - name: markdownlint
+    branches:
+    - main
     run_if_changed: '(\.md|markdownlint\.sh)$'
     decorate: true
     spec:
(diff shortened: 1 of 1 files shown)


tuminoid:
Your thoughts @lentzi90 ? Can we somehow add a check to "check-prow-config" that all tests need to have branch defined? Otherwise if we add something new here and forget the branches, it'll again go after all branches in that repo....

lentzi90:
I didn't manage to find anything for checkconfig, but I found out that it is possible to explicitly set the branchprotector to NOT manage branches. See https://docs.prow.k8s.io/docs/components/optional/branchprotector/#scope
What if we set it by default to not protect at the org/repo level and then override the specific repos/branches that we want to protect? Do you think that will work?

If I understand the current config correctly, we say that all branches in all repos under Nordix and metal3-io should be protected. At least for Nordix I don't think that is necessary. It should be enough to specify some repos (dev-tools?).

tuminoid:
> I didn't manage to find anything for checkconfig, but I found out that it is possible to explicitly set the branchprotector to NOT manage branches. See https://docs.prow.k8s.io/docs/components/optional/branchprotector/#scope What if we set it by default to not protect at the org/repo level and then override the specific repos/branches that we want to protect? Do you think that will work?
> 
> If I understand the current config correctly, we say that all branches in all repos under Nordix and metal3-io should be protected. At least for Nordix I don't think that is necessary. It should be enough to specify some repos (dev-tools?).

We will definitely need some way to specify what should be protected, namely main + release-* branches. I'll try the whitelisting instead, as even the config in this PR did not prevent branchprotector wandering off to all branches, and it would also be really sensitive breaking again if any test without branches listed explicily would cause it to go haywire again.

tuminoid:
/test metal3-ubuntu-e2e-integration-test-main 
Checking the CI, ignore the result.

tuminoid:
For now, the branchprotector job will not be added due bugs/shortcomings.

/close

PR #749, opened 2024-05-10
Fix git credentials ID

lentzi90:
We have been using the wrong credentials ID, resulting in "not found" errors (but still working without credentials). The thing is that the git plugin only supports username/password, NOT token. We have been trying to use a token for checking out the tested repo. We have been using the correct credentials for checking out project-infra though.

This PR changes all pipelines to use the same credentials ID we always used for checking out project-infra.


lentzi90:
/hold
Let's wait for next week before we merge this.
/test metal3-ubuntu-e2e-integration-test-main

furkatgofurov7 (approved):
/approve

Sunnatillo (commented):
/lgtm

PR #756, opened 2024-05-14
Prow: Enable triggering of GitHub workflows

lentzi90:
This should allow Prow to trigger (and re-trigger) workflows, making it possible for non-admins to re-run failed workflows.

### DIFF
--- a/prow/manifests/overlays/metal3/plugins.yaml
+++ b/prow/manifests/overlays/metal3/plugins.yaml
@@ -152,3 +152,7 @@ require_matching_label:
 
 override:
   allow_top_level_owners: true
+
+triggers:
+- # Enable prow to trigger GitHub workflows.
+  trigger_github_workflows: true


lentzi90:
The GitHub token that Prow uses should already have all the necessary privileges. Not sure if there is more we could test before merging this?

tuminoid:
Does workflow need to have `workflow_call` or `workflow_dispatch` so they can be "manually" triggered by Prow?  Usually triggers only obey the triggers defined by `on:`, and for example in the OSV scanner workflow we have `workflow_call` enabled so it can be triggered by another workflow.

lentzi90:
Good question! I will try to find out

lentzi90:
Looks like no special config is needed for the workflows. For example CAPI is [using this functionality](https://github.com/kubernetes/test-infra/pull/25736) and has [this workflow](https://github.com/kubernetes-sigs/cluster-api/blob/main/.github/workflows/pr-golangci-lint.yaml) without `workflow_call` and `workflow_dispatch`. They did have [some issues](https://github.com/kubernetes/test-infra/issues/25831#issuecomment-1589005165) but not related to the workflow from what I understand.

tuminoid:
There is hilarious lack of documentation about this. Searching the config option almost gives you the rare Google unicorn (1 hit) search result :)

lentzi90:
> There is hilarious lack of documentation about this. Searching the config option almost gives you the rare Google unicorn (1 hit) search result :)

Yep! I feel the need to start contributing to the Prow documentation :man_technologist: :book:

tuminoid (commented):
Even if there is zero documentation, but it is used by others, let's take it and see how it performs. 

/approve

PR #758, opened 2024-05-16
Prow: Add metal3-io org to triggers config

lentzi90:
This should hopefully make it possible to retrigger GitHub workflows on all repos in metal3-io.

### DIFF
--- a/prow/manifests/overlays/metal3/plugins.yaml
+++ b/prow/manifests/overlays/metal3/plugins.yaml
@@ -1,3 +1,4 @@
+# See https://github.com/kubernetes-sigs/prow/blob/main/pkg/plugins/plugin-config-documented.yaml
 owners:
   # By default, prow will use github org membership as the list of who is allowed to
   # /lgtm PRs.  This configuration setting tells it to use the OWNERS file only and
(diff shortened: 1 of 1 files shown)


lentzi90:

See example here: https://github.com/kubernetes-sigs/prow/blob/dd5d0eeaa28f8ec693371955e01b5d9b55ad8ad9/pkg/plugins/plugin-config-documented.yaml#L598-L624

I now added the whole metal3-io org. Do we want individual repos instead?

prow/manifests/overlays/metal3/plugins.yaml
```diff
@@ -154,5 +155,7 @@ override:
   allow_top_level_owners: true
 
 triggers:
-- # Enable prow to trigger GitHub workflows.
+- repos:
+  - metal3-io
```
tuminoid, line 159: Can you put entire org there? In test-infra there is always org/repo pair, and the config is named `repos:` which everywhere else means specfic repository?

/hold

tuminoid: Ah, nevermind. It specifically says "or only org" in the example config.

/unhold
/lgtm

tuminoid:
This is now enabling workflow triggering via `/retest` as seen linked PR above. :tada:

PR #761, opened 2024-05-17
add OWNERS_ALIASES support

tuminoid:
Add OWNERS_ALIASES file so we can reuse the groups. This way we can assign maintainers as reviewers, so blunderbuss will take both approvers and reviewers into question of picking relevant reviewers. In most of Metal3 repos, we have same or more approvers than reviewers, which causes all the review requests currently to pile up to the same 2-3 people, while approvers get none.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,16 +1,11 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 approvers:
-- kashifest
-- lentzi90
-- mboukhalfa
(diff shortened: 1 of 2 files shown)


Rozzii:
e2e is not needed

Sunnatillo (commented):
/lgtm

Sunnatillo (commented):
/approve

PR #764, opened 2024-05-20
add cherrypicker to sles-ironic-python-agent-builder

tuminoid:
It has two branches, enable cherrypicker for it.


kashifest (commented):
/approve

Issue #765, opened 2024-05-20
Rotate the dev key in DiB image building workflow


NymanRobin:
When building an image with DiB it accepts a environment variable called  `DIB_DEV_USER_AUTHORIZED_KEYS`. This takes a file path and this file will be copied to created image as the authorized_key file for the user defined in `DIB_DEV_USER_USERNAME` environment variable.

To rotate the current key a new ed25519 key should be generated and added to the authorized key file to be accepted to login with. Once the new key is validated to be working the old key should be rotated out and removed

PR #766, opened 2024-05-20
Add new key to authorized_keys in the DiB build

NymanRobin:
This is part of rotating keys we will eventually remove the old key once we have verified everything works with the new key.
The new key details are available in the password manager. 

Fixes: #765


smoshiur1237 (commented):
/lgtm

kashifest (commented):
/approve

PR #768, opened 2024-05-22
Build diskimage builder from upstream commit until next release

Sunnatillo:
Centos image building is broken with wrong partition UUID error. Here is the[ fix in the upstream](https://github.com/openstack/diskimage-builder/commit/c03e46d9e1244d6ee2ac85ff119fed54b7e08c0e). 
Until diskimage-builder is released, we need to built it from fix commit in master branch.

### DIFF
--- a/jenkins/image_building/build-image.sh
+++ b/jenkins/image_building/build-image.sh
@@ -13,11 +13,27 @@ source "${current_dir}/upload-ci-image.sh"
 source "${current_dir}/upload-node-image.sh"
 
 # Disable needrestart interactive mode
-sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null
+sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null || true
 
 sudo apt-get update
 sudo apt-get install python3-pip qemu qemu-kvm -y
 sudo pip3 install diskimage-builder python-openstackclient
+# TODO(Sunnatillo): When newer version than 3.32.0 of disk-image builder released remove changes
+# done by this commit. 
+sudo -H pip install virtualenv
+
+# sudo pip3 install diskimage-builder python-openstackclient
+mkdir "${current_dir}/dib"
+pushd "${current_dir}/dib"
+virtualenv env
+# shellcheck disable=SC1091
+source env/bin/activate
+git clone https://opendev.org/openstack/diskimage-builder
+cd diskimage-builder
+git checkout 4d1e1712b1448b12b97780c4b4cd962646884abb
+sudo pip install --no-cache-dir -e .
+
+popd
 
 export ELEMENTS_PATH="${current_dir}/dib_elements"
 export DIB_DEV_USER_USERNAME="metal3ci"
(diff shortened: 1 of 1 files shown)


jenkins/image_building/build-image.sh
```diff
@@ -13,11 +13,26 @@ source "${current_dir}/upload-ci-image.sh"
...
+# done by this commit. 
+sudo -H pip install virtualenv
+
+# sudo pip3 install diskimage-builder python-openstackclient
+mkdir "${current_dir}/dib"
+cd "${current_dir}/dib"
+virtualenv env
+# shellcheck disable=SC1091
+source env/bin/activate
+git clone https://opendev.org/openstack/diskimage-builder
```
NymanRobin, line 31: First off great that you fixed this but some thoughts that popped in my mind, not 100% necessary
 
Maybe checkout the specific commit you verified as there is some concerns regarding stability and security when just taking master without knowing what has changed there since last time?

The other option we could do here is simply build the package upload as a tarball to artifactory and then pip install directly from there as pip allows installations from urls. Then it would be quite minimal temporary change at least but, I guess that is mostly cosmetics

NymanRobin (commented):
Awesome!
/approve

jenkins/image_building/build-image.sh
Sunnatillo: Thank you for the review. Valid point. I changed it to specific commit.

jenkins/image_building/build-image.sh
```diff
@@ -13,11 +13,27 @@ source "${current_dir}/upload-ci-image.sh"
...
+cd diskimage-builder
+git checkout 4d1e1712b1448b12b97780c4b4cd962646884abb
+sudo pip install -e .
```
tuminoid, line 34:
```suggestion
sudo pip install --no-cache-dirl -e .
```
Use `--no-cache-dir` to avoid building caches outside venv.

Sunnatillo: Thank you for the review. I addressed all the comments.

jenkins/image_building/build-image.sh
```diff
@@ -13,11 +13,27 @@ source "${current_dir}/upload-ci-image.sh"
...
+
+cd ../..
```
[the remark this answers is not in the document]
Sunnatillo: done

jenkins/image_building/build-image.sh
```diff
@@ -13,11 +13,27 @@ source "${current_dir}/upload-ci-image.sh"
...
 sudo apt-get update
 sudo apt-get install python3-pip qemu qemu-kvm -y
 sudo pip3 install diskimage-builder python-openstackclient
+# TODO(Sunnatillo): When newer version than 3.32.0 of disk-image builder released remove changes
+# done by this commit. 
+sudo -H pip install virtualenv
+
+# sudo pip3 install diskimage-builder python-openstackclient
+mkdir "${current_dir}/dib"
+cd "${current_dir}/dib"
```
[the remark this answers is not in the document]
Sunnatillo: done

smoshiur1237 (commented):
Thanks for this addition
/lgtm

tuminoid (commented):
/approve

/hold
Unhold when you've manually verified the image with the latest changes.

Sunnatillo:
manually tested here:
https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3_periodic_node_image_building/23/consoleFull

PR #770, opened 2024-05-27
Prow: Enable reporting again in jenkins-operator

lentzi90:
Disabling the reporting did get rid of the double failure messages, but it also broke updates to the status contexts. They only updated after jobs completed or not at all.
Therefore we enable reporting again.


lentzi90:
/cc @kashifest 

This is already applied since at least a couple of weeks. I forgot to push a PR for it.

Sunnatillo (commented):
/approve

PR #778, opened 2024-06-04
Make BMO e2e tests optional

lentzi90:
We are moving the GitHub workflows so we should no longer require the jenkins job. Still leaving it as optional though so we can trigger it if wanted.


lentzi90:
/cc @kashifest 
Please also update the branch protection to match this!

smoshiur1237 (commented):
/lgtm

Sunnatillo (commented):
/approve

PR #788, opened 2024-06-14
Fix manifest collection in failed test

smoshiur1237:
This will add the capability to add manifest in failed test and will improve the debugging capability.

### DIFF
--- a/jenkins/scripts/files/run_fetch_logs.sh
+++ b/jenkins/scripts/files/run_fetch_logs.sh
@@ -15,10 +15,6 @@ else
 fi
 mkdir -p "${LOGS_DIR}"
 
-# Fetch cluster manifests
-mkdir -p "${LOGS_DIR}/manifests"
-cp -r /tmp/manifests/* "${LOGS_DIR}/manifests"
-
 # Fetch target cluster logs which were collected before re-pivoting
 if [[ -d "/tmp/target_cluster_logs" ]]; then
     mkdir -p "${LOGS_DIR}/k8s_target_cluster"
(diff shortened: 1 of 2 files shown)


jenkins/scripts/files/run_fetch_logs.sh
```diff
@@ -58,6 +54,57 @@ fetch_k8s_logs()
...
+      m3dataclaim
+      m3datatemplate
+    )
+
+    NAMESPACES="$(kubectl --kubeconfig="${kconfig}" get namespace -o jsonpath='{.items[*].metadata.name}')"
+    for NAMESPACE in ${NAMESPACES}; do
+      for kind in "${manifests[@]}"; do
+        mkdir -p "${LOGS_DIR}/${dir_name}/${kind}"
+        for name in $(kubectl --kubeconfig="${kconfig}" get -n "${NAMESPACE}" -o name "${kind}" || true); do
+          kubectl --kubeconfig="${kconfig}" get -n "${NAMESPACE}" -o yaml "${name}" | tee "${LOGS_DIR}/${dir_name}/${kind}/$(basename "${name}").yaml" || true
```
mboukhalfa, line 92: if we have the same kind in different namespace then would those get overridden ?

tuminoid: Good catch, and we probably have the same issue in `fetch_manifests.sh` where this was copied from. Namespace must be part of the filename or path, or you need to fetch all at once, not per namespace.

/hold

smoshiur1237: If I use the Namespace as part of the path, that should solve our issue, isn't it?

tuminoid: You need to put the namespace to the `mkdir` as well.

tuminoid:
/test metal3-ubuntu-e2e-integration-test-main metal3-centos-e2e-integration-test-main 

One of them should pass right now, and another fail, so it'd be a good test.

tuminoid (commented):
/approve

Should be fine, let's check the tests logs when completed.

tuminoid:
> /approve
> 
> Should be fine, let's check the tests logs when completed.

No manifests in the centos failure case? https://jenkins.nordix.org/view/Metal3/job/metal3-centos-e2e-integration-test-main/329/

smoshiur1237:
> > /approve
> > Should be fine, let's check the tests logs when completed.
> 
> No manifests in the centos failure case? https://jenkins.nordix.org/view/Metal3/job/metal3-centos-e2e-integration-test-main/329/

Yes I know the reason. The file I have changed was for dev-env ansible test run. For e2e We are using the dynamic worker workflow where it has the same fetch log for e2e test. So I will update this PR to add the same changes in dynamic worker workflow

Rozzii (commented):
/lgtm

PR #797, opened 2024-06-26
:seedling: Use basic-checks image instead of docker.io/golang for basic tests in BMO repo

mquhuy:
As mentioned by https://github.com/metal3-io/project-infra/pull/796, this PR is to replace `docker.io/golang` with `quay.io/metal3-io/basic-checks` image

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -393,7 +393,7 @@ presubmits:
         env:
         - name: IS_CONTAINER
           value: "TRUE"
-        image: docker.io/golang:1.22
+        image: quay.io/metal3-io/basic-checks:golang-1.22
         imagePullPolicy: Always
   - name: gomod
     branches:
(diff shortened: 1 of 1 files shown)


tuminoid (commented):
Pulling the image without logging in doesn't work, so I think this would fail right now.

tuminoid:
> Pulling the image without logging in doesn't work, so I think this would fail right now.

Actually the image has not been pushed yet, hence the failure, but even more so /hold is in order.

mquhuy:
> > Pulling the image without logging in doesn't work, so I think this would fail right now.
> 
> Actually the image has not been pushed yet, hence the failure, but even more so /hold is in order.

The image is pushed. It's in https://quay.io/repository/metal3-io/basic-checks?tab=tags

tuminoid:
Wouldn't it have been easier to just publish the image? There is nothing secret about it, and for private images you need to worry about tests in `hack/` directories as well eventually etc.

mquhuy:
> Wouldn't it have been easier to just publish the image? There is nothing secret about it, and for private images you need to worry about tests in `hack/` directories as well eventually etc.

Ah. Silly me, I couldn't find a way to make it public. Just now I found it :fearful:

tuminoid:
As agreed offline, only apply the new image to BMO and only where it is required. Also make PR to BMO to make `hack/generate.sh` use the same image, otherwise local runs of generate test will fail too.

tuminoid (commented):
/unhold
/approve

Sunnatillo (commented):
/lgtm

PR #784, opened 2024-06-11
Push CI images to both Xerces and Cleura

kashifest:
This PR updates image upload to both xerces and Cleura. 

Signed-off-by: Kashif Khan <kashif.khan@est.tech>

PR #799, opened 2024-06-28
delete local images after they're uploaded

tuminoid:
Delete local images after they're uploaded as they're huge and might cause disk space issues if dynamic worker is reused.

Reference: https://github.com/metal3-io/project-infra/pull/784


Sunnatillo (commented):
/approve

kashifest (commented):
/lgtm

PR #808, opened 2024-07-03
Remove old SSH key from the CI image

kashifest:
This is part of the CI key rotation. The new key seems to work in most of the CI jobs. This PR gets rid of the old one.


tuminoid (commented):
/approve

We should stop injecting keys like this.

kashifest:
> We should stop injecting keys like this.

What are the alternatives to this if we want to have a key in the CI image?

tuminoid:
> > We should stop injecting keys like this.
> 
> What are the alternatives to this if we want to have a key in the CI image?

Maybe inject via JJB from Jenkins config or download it during build. We'd need to get in position to do the rotation in 1 day, preferably automated so PRs work badly for that.

PR #811, opened 2024-07-10
Run feature tests independently

smoshiur1237:
This PR will allow to run feature tests in three different parts such as pivoting based, remediation based and other features(non pivot scenario). So it will help to run them independently and reduce resource consumption in case of re-run.


jenkins/jobs/e2e_features_test.pipeline
```diff
@@ -0,0 +1,120 @@
...
+    IMAGE_OS = "${IMAGE_OS}"
+    CAPI_VERSION = "${CAPI_VERSION}"
+    CAPM3_VERSION = "${CAPM3_VERSION}"
+    CAPM3RELEASEBRANCH = "${capm3_release_branch}"
+    BMORELEASEBRANCH = "${bmo_release_branch}"
+    TARGET_NODE_MEMORY = "${TARGET_NODE_MEMORY}"
+    NUM_NODES = 4
+    TESTS_FOR = "${TESTS_FOR}"
+    GINKGO_FOCUS = "${GINKGO_FOCUS}"
+    GINKGO_SKIP = "${GINKGO_SKIP}"
```
smoshiur1237, line 57: I need suggestion: Do I need to add `GINKGO_SKIP` variable in pipeline? JJB is not going to pass this var but this pipeline will add the value

smoshiur1237:
This PR needs to get merge same time with [JJB PR](https://gerrit.nordix.org/c/infra/cicd/+/21725)

smoshiur1237:
@Sunnatillo  would you please override the required test?

jenkins/jobs/e2e_features_test.pipeline
Rozzii: If JJB is not passing it I don't see much use of this var in this form. This can be either removed or if you need you can hardcode some value here e.g. GINKGO_SKIP="whatever"

Sunnatillo: it is necessary to pass GINKGO_SKIP when GINKGO_FOCUS=features. 
You can add condition here in the pipeline  GINKGO_FOCUS == features, GINKGO_SKIP== pivoting  remediation.

Rozzii: Okay you are giving a value here https://github.com/metal3-io/project-infra/pull/811/files#diff-7654f3f0b93e2e0e5eac4e5b579b787cc7644bd0ab047192211c8dbc5d785cfeR33

Rozzii (commented):
/lgtm

jenkins/jobs/e2e_features_test.pipeline
smoshiur1237: After discussion, we can see that this var is needed

Sunnatillo (approved):
/approve

PR #813, opened 2024-07-12
:seedling: Set trap to cleanup in build-image.sh

mquhuy:
In node image building script, if a failure happens in the middle of the script, the node image and the `dib` directory should be cleaned, so that the next pipeline can start.

For e.g. in https://jenkins.nordix.org/job/metal3_periodic_node_image_building/32/execution/node/62/log/
```
+ mkdir /home/****/workspace/metal3_periodic_node_image_building/jenkins/image_building/dib
mkdir: cannot create directory ‘/home/****/workspace/metal3_periodic_node_image_building/jenkins/image_building/dib’: File exists
```

### DIFF
--- a/jenkins/image_building/build-image.sh
+++ b/jenkins/image_building/build-image.sh
@@ -2,10 +2,18 @@
 
 set -eux
 
+current_dir="$(dirname "$(readlink -f "${0}")")"
+
+cleanup() {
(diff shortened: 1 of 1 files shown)


Rozzii (commented):
/lgtm
Thanks, nice improvement!

PR #816, opened 2024-07-16
:bug: Fix IMAGE_LOCATION in build-image.sh

mquhuy:
[node_image_building](https://jenkins.nordix.org/job/metal3_periodic_node_image_building/) is currently failing because the location of the file is incorrect. The built image locates in the directory, from where the script is run (which is the repo root), while the verify script assumes that image locates in `jenkins/image_building`.

This PR fixes the IMAGE_LOCATION, so that dev-env could find the image.

Edit: Also add the `${IMAGE_NAME}.d` directory cleanup. This directory was created as a part of the image building.

### DIFF
--- a/jenkins/image_building/build-image.sh
+++ b/jenkins/image_building/build-image.sh
@@ -3,14 +3,19 @@
 set -eux
 
 current_dir="$(dirname "$(readlink -f "${0}")")"
+REPO_ROOT="$(realpath "${current_dir}/../..")"
 
 cleanup() {
   deactivate || true
-  sudo rm -f "${img_name}".{raw,qcow2}
+  sudo rm -f "${REPO_ROOT}/${img_name}".{raw,qcow2}
+  sudo rm -rf "${REPO_ROOT}/${img_name}.d"
   sudo rm -rf "${current_dir}/dib"
 }
 
 trap cleanup EXIT
+
+# Make sure we run everything in the repo root
+cd "${REPO_ROOT}" || true
 
 export IMAGE_OS="${IMAGE_OS}"
 export IMAGE_TYPE="${IMAGE_TYPE}"
@@ -76,7 +81,7 @@ disk-image-create --no-tmpfs -a amd64 -o "${img_name}".qcow2 "${IMAGE_OS}"-"${IM
 
 
 if [[ "${IMAGE_TYPE}" == "node" ]]; then
-  verify_node_image "${img_name}"
+  verify_node_image "${img_name}" "${REPO_ROOT}"
   echo "Image testing successful."
   upload_node_image "${img_name}"
 else
--- a/jenkins/image_building/verify-node-image.sh
+++ b/jenkins/image_building/verify-node-image.sh
@@ -3,9 +3,10 @@
 set -eux
 
 verify_node_image() {
-    img_name="$1"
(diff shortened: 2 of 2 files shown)


jenkins/image_building/build-image.sh
```diff
@@ -3,15 +3,21 @@
 set -eux
 
 current_dir="$(dirname "$(readlink -f "${0}")")"
+REPO_ROOT="$(realpath "${current_dir}/../..")"
 
 cleanup() {
   deactivate || true
+  cd "${REPO_ROOT}" || true
   sudo rm -f "${img_name}".{raw,qcow2}
+  sudo rm -rf "${img_name}.d"
```
mboukhalfa, line 12: I do not understand this line ! Is it deleting a directory or an extension of the image

mquhuy: There's an extra directory created by diskimage-builder and needs to be cleaned as well. Sorry I missed that in the PR description.

jenkins/image_building/build-image.sh
```diff
@@ -3,15 +3,21 @@
 set -eux
 
 current_dir="$(dirname "$(readlink -f "${0}")")"
+REPO_ROOT="$(realpath "${current_dir}/../..")"
 
 cleanup() {
   deactivate || true
+  cd "${REPO_ROOT}" || true
```
Rozzii, line 10: No need to use `cd` here.
  sudo rm -f "${REPO_ROOT}/${img_name}".{raw,qcow2}
  sudo rm -rf "${REPO_ROOT}/${img_name}.d"

jenkins/image_building/build-image.sh
```diff
@@ -3,15 +3,21 @@
...
   sudo rm -f "${img_name}".{raw,qcow2}
+  sudo rm -rf "${img_name}.d"
   sudo rm -rf "${current_dir}/dib"
 }
 
 trap cleanup EXIT
 
+# Make sure we run everything in the repo root
+cd "${REPO_ROOT}" || true
```
Rozzii, line 19: cd is fine here you are not expecting to leave this location.

Rozzii (changes requested):
minor nit

jenkins/image_building/build-image.sh
mquhuy: Good catch! Thanks

jenkins/image_building/build-image.sh
mquhuy: Thanks. But I still would like to keep this!

One major problem with this script (and with many of our scripts), is that we always assume that the script starts from some location, in this case the REPO_ROOT. It works fine for CI, but in case one wants to run it from somewhere else (like cd jenkins/image_building && ./build-image.sh), it wouldn't work. This cd will help solve that issue, and it's harmless otherwise.

Sunnatillo:
@mquhuy, now I remember one issue I faced. 
DiskImageBuilder also has a variable called IMAGE_LOCATION.  This might be conflicting with dev-env's IMAGE_LOCATION variable,

jenkins/image_building/build-image.sh
Rozzii: Yes I agree that is why I said it is fine :) .

Rozzii (approved):

PR #820, opened 2024-07-19
:seedling: Set trap to run `make clean` after integration test

mquhuy:
Add an option to cleanup the system (by running dev-env `make clean`) after integration test. This is to make sure the next pipeline can start from a clean state, hence avoid issues like the one we see in https://jenkins.nordix.org/job/metal3_periodic_node_image_building/38/


Rozzii (commented):
/lgtm

PR #825, opened 2024-07-26
Test capm3 release 1.6 with golang 1.22

NymanRobin:
Since golang 1.20 is already deprecated

This is needed for PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/1858

### DIFF
--- a/prow/manifests/overlays/metal3/config.yaml
+++ b/prow/manifests/overlays/metal3/config.yaml
@@ -1022,6 +1022,7 @@ presubmits:
     branches:
     - main
     - release-1.7
+    - release-1.6
     skip_if_only_changed: '(((^|/)OWNERS)|((^|/)OWNERS_ALIASES)|(\.md))$'
     decorate: true
     spec:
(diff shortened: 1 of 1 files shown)


NymanRobin:
Another note here is that the release 1.5 could be removed from the config to reduce the already long config? Since the branch is not supported anymore

adilGhaffarDev:
> Another note here is that the release 1.5 could be removed from the config to reduce the already long config? Since the branch is not supported anymore

CI is still running for 1.5. Will it effect CI? I would suggest wait for one week, only a week is left after that we are removing ci jobs of 1.5.

adilGhaffarDev (commented):
/lgtm

NymanRobin:
> > Another note here is that the release 1.5 could be removed from the config to reduce the already long config? Since the branch is not supported anymore
> 
> CI is still running for 1.5. Will it effect CI? I would suggest wait for one week, only a week is left after that we are removing ci jobs of 1.5.

Yes it would effect CI, so better wait :smile: Thanks for the clarification though

Sunnatillo (commented):
/approve

PR #827, opened 2024-07-30
:seedling: Change image-building pipeline to a matrix

mquhuy:
This is to split ubuntu and centos image building into separate VMs and have proper indication for each step results in Jenkins UI.

### DIFF
--- a/jenkins/image_building/upload-node-image.sh
+++ b/jenkins/image_building/upload-node-image.sh
@@ -68,3 +68,8 @@ upload_node_image() {
         echo "${MAPFILE[i]} has been deleted!"
     done
 }
+
+# If the script was run directly (i.e. not sourced), run the upload_node_image func
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+    upload_node_image "$@"
+fi
--- a/jenkins/jobs/image_building.pipeline
+++ b/jenkins/jobs/image_building.pipeline
@@ -3,8 +3,7 @@ ci_git_branch = "main"
 ci_git_url = "https://github.com/metal3-io/project-infra.git"
 
 pipeline {
-  agent { label 'metal3ci-jenkins' }
-  options { ansiColor('xterm') }
+  agent none
   environment {
     IMAGE_TYPE = "${IMAGE_TYPE}"
     METAL3_CI_USER="metal3ci"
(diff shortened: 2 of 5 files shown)


jenkins/jobs/image_building.pipeline
```diff
@@ -15,61 +14,88 @@ pipeline {
...
-          usernamePassword(credentialsId: 'metal3ci_city_cloud_openstack_credentials', usernameVariable: 'OPENSTACK_USERNAME_CLEURA', passwordVariable: 'OPENSTACK_PASSWORD_CLEURA'),
-          usernamePassword(credentialsId: 'infra-nordix-artifactory-api-key', usernameVariable: 'RT_USER', passwordVariable: 'RT_TOKEN')
-        ]) {
-          catchError([stageResult: 'FAILURE', message: "Failed to build Centos CI image"]) {
-            sh "./jenkins/image_building/build-image.sh"
+        environment {
+          IMAGE_OS = "${IMAGE_OS}"
+        }
+        stages {
+          stage("Checkout CI Repo") {
```
Sunnatillo, line 30: Can we merge stage("Checkout CI Repo") and stage("Build CI image"). Checking out does not necessarily need to have its own stage. We are having too many granular steps on the view.  

image

jenkins/image_building/upload-node-image.sh
```diff
@@ -68,3 +68,7 @@ upload_node_image() {
         echo "${MAPFILE[i]} has been deleted!"
     done
 }
+
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
```
tuminoid, line 72: This is unintuitive for readers, put a comment there what it does and when it is triggered

jenkins/jobs/image_building.pipeline
```diff
@@ -1,10 +1,9 @@
 ci_git_credential_id = "metal3-jenkins-github-username-token"
-ci_git_branch = "main"
-ci_git_url = "https://github.com/metal3-io/project-infra.git"
+ci_git_branch = "mquhuy/image-build-matrix"
```
tuminoid, line 2: Not mergerable with this, so
/hold

jenkins/jobs/image_building.pipeline
mquhuy: I have no big issue, but I tend towards disagreeing with that. We've most likely always put the checkout as a step in all of our jobs, I'd prefer to keep it that way for consistency.

jenkins/image_building/upload-node-image.sh
mquhuy: Thanks. I thought that was a common practice, but I could be wrong.

tuminoid (commented):
/approve
/cc @Sunnatillo @kashifest

smoshiur1237 (commented):
/lgtm

PR #833, opened 2024-08-07
:seedling: Add container image build wf template

mquhuy:
This template is not meant to be used in this repo, but to be referenced by container build wf from other repos.

An example of how it could be used: https://github.com/metal3-io/baremetal-operator/actions/runs/10280659565/job/28448467379?pr=1861

### DIFF
--- a/.github/workflows/container-image-build.yml
+++ b/.github/workflows/container-image-build.yml
@@ -1,0 +1,77 @@
+name: build-images action template
+permissions: {}
+
+on:
+  workflow_call:
+    inputs:
+      image-name:
+        required: true
+        description: "Name of the image to build"
+        type: string
+      dockerfile-directory:
+        required: false
+        description: "The directory where the dockerfile locates, as relative to the repo root"
+        type: string
+        default: .
+      pushImage:
+        required: false
+        description: "Whether to push the image afterwards"
+        type: boolean
+        default: true
+    secrets:
+      QUAY_USERNAME:
+        required: true
+      QUAY_PASSWORD:
+        required: true
+      SLACK_WEBHOOK:
+        required: true
+
+jobs:
+  job:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'metal3-io'
+
+    permissions:
+      contents: write
+
+    steps:
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      with:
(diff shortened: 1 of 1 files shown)


tuminoid:
Can you already test this, PR + PR combo?

mquhuy:
> Can you already test this, PR + PR combo?

Yes. I have tested it, but I modified the BMO PR now to illustrate how the test is: https://github.com/metal3-io/baremetal-operator/pull/1861

tuminoid (commented):
/approve

I think we can merge this already, as its not doing anything on its own. It is to be called by other repos, starting with BMO and then all the other repos converted to this.

.github/workflows/container-image-build.yml
```diff
@@ -0,0 +1,70 @@
...
+      pushImage:
+        required: false
+        description: "Whether to push the image afterwards"
+        type: boolean
+        default: true
+
+jobs:
+  job:
+    runs-on: ubuntu-latest
+    if: github.repository.fork == false
```
tuminoid, line 25: Just a thought: if this is not specifically checking for metal3-io/project-infra, can any non-fork repository call this workflow and push stuff with it? Why are checking for forks only?

/hold

mquhuy: It shouldn't check for `metal3-io/project-infra`, because the `github.repository` will be the caller's repository (`metal3-io/baremetal-operator`, for e.g.)

Technically any repo can call this template, but the secrets from this repo won't be exposed to anyone. Afaiu, workflow template works just like a go/python package, any script can import it, but the caller needs to provide the necessary environment for it to run.

tuminoid: OK, thanks for clarifying it. Makes sense, but you cannot always be sure with Github :)

tuminoid: /unhold

mquhuy: @tuminoid I changed the logic now, so that it will only be triggered by repos inside `metal3-io`. Please notice as well that `project-infra` repo doesn't have access to the newly-added secrets `QUAY_USERNAME` and `QUAY_PASSWORD`.

tuminoid:
So we changed from username + token to username + password? Is this tested now?

tuminoid:

/cc @kashifest 
Please lgtm this.

kashifest (commented):
/lgtm

PR #840, opened 2024-08-20
:seedling: Add python3-dev package to build-image.sh

mquhuy:
This helps solve the ongoing failure in [node-image-building job](https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3_periodic_node_image_building/)

https://jenkins.nordix.org/blue/organizations/jenkins/metal3_periodic_node_image_building/detail/metal3_periodic_node_image_building/78/pipeline/

In general, this PR improves the script by:
- Removing global scope pip install (`sudo pip3 install`), which is not recommended by PEP8, and will stop working in later Python version.
- Removing unnecessary installation of package `python-openstackclient`
- Removing unnecessary clone of `diskimage-builder` repo
- Using `python-venv` (a python standard library) instead of `virtualenv` (package from PyPI).
- Minor syntax updates.

### DIFF
--- a/jenkins/image_building/build-image.sh
+++ b/jenkins/image_building/build-image.sh
@@ -8,7 +8,7 @@ REPO_ROOT="$(realpath "${current_dir}/../..")"
 cleanup() {
   deactivate || true
   sudo rm -rf "${REPO_ROOT}/${img_name}.d"
-  sudo rm -rf "${current_dir}/dib"
+  sudo rm -rf "${current_dir}/venv"
 }
 
 trap cleanup EXIT
@@ -20,29 +20,18 @@ export IMAGE_OS="${IMAGE_OS}"
 export IMAGE_TYPE="${IMAGE_TYPE}"
 
 # shellcheck disable=SC1091
-source "${current_dir}/upload-ci-image.sh"
+. "${current_dir}/upload-ci-image.sh"
 
 # Disable needrestart interactive mode
 sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null || true
 
 sudo apt-get update
-sudo apt-get install python3-pip qemu qemu-kvm -y
-sudo pip3 install diskimage-builder python-openstackclient
-sudo -H pip install virtualenv
+sudo apt-get install -y python3-dev python3-pip python3-venv qemu qemu-kvm
+python3 -m venv venv
 
-# sudo pip3 install diskimage-builder python-openstackclient
-mkdir "${current_dir}/dib"
-pushd "${current_dir}/dib"
-virtualenv env
-# shellcheck disable=SC1091
-source env/bin/activate
-
-git clone https://opendev.org/openstack/diskimage-builder || true
-cd diskimage-builder
-git checkout 3.33.0
-sudo pip install --no-cache-dir -e .
-
-popd
+# shellcheck source=/dev/null
+. venv/bin/activate
+pip install --no-cache-dir diskimage-builder==3.33.0
 
 export ELEMENTS_PATH="${current_dir}/dib_elements"
 export DIB_DEV_USER_USERNAME="metal3ci"


smoshiur1237 (commented):
/lgtm

Sunnatillo:
I used virtual env to install specific version from github when latest version had an issues. Now it is resolved. Do we still need this? We can use latest version.

mquhuy:
> I used virtual env to install specific version from github when latest version had an issues. Now it is resolved. Do we still need this? We can use latest version.

IMO venv is slightly better as it's the standard package. I wouldn't  oppose to virtualenv either, but was there any specific reason for using it in the first place? If not, I would say using venv is better.

Btw, if the issue was resolved, how come the job still failed?

jenkins/image_building/build-image.sh
```diff
@@ -26,21 +26,20 @@ source "${current_dir}/upload-ci-image.sh"
 sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null || true
 
 sudo apt-get update
-sudo apt-get install python3-pip qemu qemu-kvm -y
-sudo pip3 install diskimage-builder python-openstackclient
-sudo -H pip install virtualenv
+sudo apt-get install python3-dev python3-pip python3-venv qemu qemu-kvm -y
```
tuminoid, line 29:
```suggestion
sudo apt-get install -y python3-dev python3-pip python3-venv qemu qemu-kvm
```

jenkins/image_building/build-image.sh
```diff
@@ -26,21 +26,20 @@ source "${current_dir}/upload-ci-image.sh"
...
+python3 -m venv venv
+
+# shellcheck source=/dev/null
+source venv/bin/activate
```
tuminoid, line 33:
```suggestion
. venv/bin/activate
```
source is deprecated

jenkins/image_building/build-image.sh
```diff
@@ -26,21 +26,20 @@ source "${current_dir}/upload-ci-image.sh"
...
+pip install diskimage-builder python-openstackclient
```
tuminoid, line 34:
```suggestion
pip install --no-cache-dir diskimage-builder python-openstackclient
```

tuminoid (commented):
Some minor nits

mquhuy:
> The issue in image building seems to be `python3-dev` missing (it cannot find `Python.h` which is coming from python3-dev), which this PR adds and hence solves the issue.

Thanks a lot. Yes, I guess you're right, the missing `python.h` only turned up when I switched to `venv`, perhaps `virtualenv` hid it in some way.

> IMO, just address the minor nits, and maybe update the PR title/description about the issue (its not venv vs virtualenv), and were good.

I will. Doing some tests now with the new code change.

jenkins/image_building/build-image.sh
```diff
@@ -20,29 +20,18 @@ export IMAGE_OS="${IMAGE_OS}"
...
 # shellcheck disable=SC1091
-source "${current_dir}/upload-ci-image.sh"
+. "${current_dir}/upload-ci-image.sh"
 
 # Disable needrestart interactive mode
 sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null || true
 
 sudo apt-get update
-sudo apt-get install python3-pip qemu qemu-kvm -y
-sudo pip3 install diskimage-builder python-openstackclient
```
tuminoid, line 30: We don't need python-openstackclient anymore?

mquhuy: No. We don't interact with openstack or ironic. Only diskimage-builder is needed.

Sunnatillo:
> > I used virtual env to install specific version from github when latest version had an issues. Now it is resolved. Do we still need this? We can use latest version.
> 
> IMO venv is slightly better as it's the standard package. I wouldn't oppose to virtualenv either, but was there any specific reason for using it in the first place? If not, I would say using venv is better.
> 
> Btw, if the issue was resolved, how come the job still failed?

The issue was long time ago. And it was resolved on 3.33.0 version. The issue we are facing is different.

tuminoid (commented):
/approve

Sunnatillo (commented):
/lgtm

mquhuy:
> The issue was long time ago. And it was resolved on 3.33.0 version. The issue we are facing is different.

Yeah, I think Tuomo was correct, it came from the missing python.h

PR #847, opened 2024-08-29
Do not upload images to fra region

Sunnatillo:
This pr removes uploading to fra region as we are moving to kna.


tuminoid (commented):
/approve
/hold
Unhold when the builds have moved.

smoshiur1237 (commented):
/lgtm

PR #854, opened 2024-09-10
Fix typo

Sunnatillo:
Fixes typo

### DIFF
--- a/jenkins/scripts/clean_resources.sh
+++ b/jenkins/scripts/clean_resources.sh
@@ -45,7 +45,7 @@ cleanup() {
 }
 
 # Install openstack client
-pip install python-openstackclient="${CLIENT_VERSION}"
+pip install python-openstackclient=="${CLIENT_VERSION}"
 
 # Cleaning up CLEURA resources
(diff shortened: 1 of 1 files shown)


jenkins/scripts/clean_resources.sh
```diff
@@ -11,7 +11,7 @@ set -eu
 # Usage:
 #  clean_resources.sh
 #
-CLIENT_VERSION="7.0.0"
+CLIENT_VERSION="v7.0.0"
```
tuminoid, line 14: Either syntax, 7.0.0 or v7.0.0 works. The version listed in PyPI is 7.0.0 so I'd use that.

jenkins/scripts/clean_resources.sh
```diff
@@ -45,7 +45,7 @@ cleanup() {
 }
 
 # Install openstack client
-pip install python-openstackclient="${CLIENT_VERSION}"
+pip install python-openstackclient=="${CLIENT_VERSION}"
```
tuminoid, line 48: This is a legit fix tho.

Sunnatillo: done

PR #858, opened 2024-09-16
Allow uploading the latest image under a common name

kashifest:
This PR makes sure that the latest image created in Openstack is pushed under a common name. This will make sure that the image being used in Jenkins workers remain active all the time. When replacing an old image with the latest, this commit also renames the old image based on its original name which is build date suffixed and this will be added in the images as an image_name property.

### DIFF
--- a/jenkins/image_building/upload-ci-image.sh
+++ b/jenkins/image_building/upload-ci-image.sh
@@ -2,6 +2,8 @@
 
 set -eux
 
+COMMON_IMAGE_NAME="metal3-ci-${IMAGE_OS}-latest"
+
 delete_old_images() {
 
   # We keep last RETENTION_NUM of metal3ci images and delete old ones when new image has been pushed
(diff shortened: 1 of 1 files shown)


jenkins/image_building/upload-ci-image.sh
```diff
@@ -68,8 +80,18 @@ upload_ci_image_xerces() {
   export OS_USER_DOMAIN_NAME="xerces"
   export OS_IDENTITY_API_VERSION=3
 
+ # Check if the common image already exists
+  if openstack image show "${COMMON_IMAGE_NAME}" &>/dev/null; then
```
Rozzii, line 84: Have you checked, does this return non zero if it can't find any image?

kashifest: Yes

tuminoid (commented):
/approve

This is great, puts stop to Jenkins worker breakages. Thanks!

PR #861, opened 2024-09-19
:seedling: Add pr-verifier-workflow

mquhuy:
This workflow will be used to replace `Kubebuilder-release-tools pr-verifier`. It can be referenced to in all of our repos.

Example of how this workflow can be used: https://github.com/metal3-io/baremetal-operator/pull/1968


Sunnatillo:
Do we need PR verifier for project-infra?

Sunnatillo (commented):
/approve

PR #870, opened 2024-10-01
fix IPA git commit/branch refference

Rozzii:
This commit:
 - Fixes an icorrect value assignment for IPA repository refference This bug was causing the build tooling to not check out the specified IPA branch instead it was checking out the mian
 - Adds missing IPA repository address variable

### DIFF
--- a/jenkins/scripts/dynamic_worker_workflow/build_ipa.sh
+++ b/jenkins/scripts/dynamic_worker_workflow/build_ipa.sh
@@ -66,7 +66,7 @@ METADATA_PATH="/tmp/metadata.txt"
 
 sudo rm -rf "${IPA_BUILD_WORKSPACE}"
 # Update apt packages
-sudo apt-get update -y 
+sudo apt-get update -y
 
 # Install required packages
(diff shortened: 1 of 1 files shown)


Sunnatillo (commented):
/approve

tuminoid:
This is not tested by integration job, but fullstack. Fullstack isn't enabled for this repo though, so we need to change prow config to add it first, then run it here.

Rozzii:
Okay it looks like metal3-ubuntu-e2e-integration-test-main  is not a relevant test here.
We have test job for testing IPA builder PRs but it is not assigned to this repo.
/hold
I will assign the test to the repo.

Rozzii:
The fullstack PR test still has the legacy ghprb trigger logic in it's pipeline code so I will upgrade it to use prow.

Rozzii:
I have thought about this and although it can be only tested by fullstack but fullstack is not yet working with prow, this can be only tested by the periodic fullstack. 

IMO this can be merged, I have tested the build script manually it does not break the build other than that it has no effect, in worst case scenario it breaks the fullstack but nothing relies on fullstack test .
/hold cancel

PR #877, opened 2024-10-15
extending the initrd_sdk

Rozzii:
This commit adds:

- New script used to avoid race conditions while running root switching
- New systemd unit file that is used to run the previously introduced unlock-mount-luks.sh
- Documentation about the new files

### DIFF
--- a/jenkins/image_building/initrd_sdk/verify-realroot.sh
+++ b/jenkins/image_building/initrd_sdk/verify-realroot.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -eu
+
+# This script is used to prvide a controlled wait loop in order to give time
+# to other systemd srvices to prepare the root file system.
+
+while true; do
+    if [[ -e "/realroot/bin" ]]; then
+        printf "INFO: Realroot mount point is present.\n"
+        break
+    else
+        printf "INFO: Waiting for realroot!\n"
+        # Introduce a 1-second delay using the read command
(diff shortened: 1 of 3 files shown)


jenkins/image_building/initrd_sdk/verify-realroot.sh
```diff
@@ -0,0 +1,19 @@
...
+while true; do
+    if [ -e "/realroot/bin" ]; then
+        printf "INFO: Realroot mount point is present.\n"
+        break
+    else
+        printf "INFO: Waiting for realroot!\n"
+        # Introduce a 1-second delay using the read command
+        # sleep might not be available but this way we stress
+        # the CPU less
+        read -r -t 1 < "/dev/null"
```
mboukhalfa, line 17: I tried this seems not waiting on my terminal ! is tested somehow ?

mboukhalfa (commented):
If I understand correctly the verify-realroot.sh is introduced but not used so far ?

jenkins/image_building/initrd_sdk/verify-realroot.sh
Rozzii: yeah I have tested it it was waiting for me although I was testing this in a initrd.

Rozzii: now it is good, I have done a modification after I have tested, nice catch.

PR #880, opened 2024-10-16
improve initrd verification logic

Rozzii:
This commit adds the following changes:

- Implements a feature for the initrd sdk that makes it possible to verify that both the root and the config drive partitions have been decrypted.
- Now the verify script is reading /dev/zero in order to generate the 1s delay, instead of reading from interactive stdin. In environments where the verify script is executed in the background or by other services e.g. cron, systemd etc... the read command returns immediately thus not providing sufficient delay.


Sunnatillo (commented):
/approve

Issue #807, opened 2024-07-03
Enable milestone and milestoneapplier plugins for Prow


tuminoid:
Enable milestone and milestoneapplier plugins for Prow to automatically assign a milestone to closed PRs, and allow milestone setting for project members in case they cannot do it via the menu.

We should grab config for this from CAPI, with the notable difference is that we have many repos, different milestones in them, and some repos do not have milestones at all.

https://prow.k8s.io/plugins

tuminoid:
@Rozzii

tuminoid:
@lentzi90 this one we discussed offline

PR #895, opened 2024-11-08
Prow: Configure milestone applier plugin

lentzi90:
This plugin will apply milstones automatically to merged PRs depending on what branch they target.

Fixes: #807

### DIFF
--- a/prow/config/plugins.yaml
+++ b/prow/config/plugins.yaml
@@ -167,3 +167,27 @@ triggers:
   - metal3-io
   # Enable prow to trigger GitHub workflows.
   trigger_github_workflows: true
+
+milestone_applier:
+  metal3-io/cluster-api-provider-metal3:
+    main: v1.9
+    release-1.8: v1.8
+    release-1.7: v1.7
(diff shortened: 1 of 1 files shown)


Sunnatillo (commented):
/approve

tuminoid:
Was the plugin already enabled, and only missed conf?

Rozzii (commented):
/lgtm

kashifest (approved):

lentzi90:
>  Was the plugin already enabled, and only missed conf?

It is one of the built-in plugins. Not sure why they even call them plugins when they are literally compiled in the same binary. In theory you could of course compile it without them but, oh well...

tuminoid:
> > Was the plugin already enabled, and only missed conf?
> 
> It is one of the built-in plugins. Not sure why they even call them plugins when they are literally compiled in the same binary. In theory you could of course compile it without them but, oh well...

OK, makes sense then why grepping for milestone never gave results in configs, and we thought it was not even enabled and needed a bit more work. Well, easy win there :)

PR #897, opened 2024-11-08
Prow: Fix milestone applier config

lentzi90:
Now it should match [the actual milestones](https://github.com/orgs/metal3-io/projects/8/views/4).

### DIFF
--- a/prow/config/plugins.yaml
+++ b/prow/config/plugins.yaml
@@ -170,24 +170,23 @@ triggers:
 
 milestone_applier:
   metal3-io/cluster-api-provider-metal3:
-    main: v1.9
-    release-1.8: v1.8
-    release-1.7: v1.7
-    release-1.6: v1.6
+    main: "CAPM3 - v1.9"
+    release-1.8: "CAPM3 - v1.8"
+    release-1.7: "CAPM3 - v1.7"
+    release-1.6: "CAPM3 - v1.6"
   metal3-io/baremetal-operator:
-    main: v0.9
-    release-0.8: v0.8
-    release-0.7: v0.7
-    release-0.6: v0.6
-    release-0.5: v0.5
+    main: "BMO - v0.9"
+    release-0.8: "BMO - v0.8"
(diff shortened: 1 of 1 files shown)


Rozzii (commented):
/lgtm

prow/config/plugins.yaml
```diff
@@ -170,24 +170,24 @@ triggers:
...
+    release-1.6: "CAPM3 - v1.6"
   metal3-io/baremetal-operator:
-    main: v0.9
-    release-0.8: v0.8
-    release-0.7: v0.7
-    release-0.6: v0.6
-    release-0.5: v0.5
+    main: "BMO - v0.9"
+    release-0.8: "BMO - v0.8"
+    release-0.7: "BMO - v0.7"
```
tuminoid, line 180: There is no BMO 0.7

tuminoid (commented):
One issue, plus older milestones still need renaming.

/hold 
until they're consistently named.

But generally, this is what we agreed, so
/approve

prow/config/plugins.yaml
Rozzii: No there was an issue during release there is a comment on the milestone-

lentzi90: Hmm but we have a branch and milestone. Should we remove it from here because we don't expect to use it or do we just keep it?

tuminoid: We should not have milestone, and branch is locked, so it should not be mentioned anywhere.

lentzi90: Makes sense. I will remove it from here.

Rozzii: We have only branch no milestone but yes IMO we can get rid of that 0,7 branch.

Rozzii (commented):
/lgtm

Rozzii:
It think this is good to go now.

lentzi90:
Thanks @Rozzii !
/hold cancel

Issue #885, opened 2024-10-28
CI image building failure during uploading the new image


smoshiur1237:
Test was failing during uploading the new image. It may be introduced by this #858, 

Test: https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3_periodic_ci_image_building/
Failure message :
`metal3-ci-ubuntu-latest BadRequestException: 400: Client Error for url: https://kna1.citycloud.com:9292/v2/images/f2abb8f0-e1bc-4260-8e9e-e2decba6fe71, Bad Request`

PR #898, opened 2024-11-14
Fix CI image upload issue

smoshiur1237:
Fixes: #885 
The issue was in the output of the `openstack image show` command, which was not having proper format that we want. So changing it to json is giving correct output.


smoshiur1237:
/cc @tuminoid @lentzi90 @Sunnatillo 
Please override the CI test.

tuminoid (commented):
/approve

Sunnatillo (commented):
/lgtm

PR #901, opened 2024-11-15
Prow: Bump calico, autoscaler and cloud-provider

lentzi90:
Also adjust minimum number of machines in the worker MD.
The cloud-provider and autoscaler bump should really have
happened together with the kubernetes bump but was forgotten.


lentzi90:
Note that we bumped kubernetes to v1.30.6 but the autoscaler and cloud-provider-openstack does not release new patches at the same time. Therefore they are on different patch versions. All have been bumped to their latest patch version within the v1.30 minor version.

tuminoid (commented):
/approve

Thanks, this will make for less autoscaling and more stable environment.

Issue #881, opened 2024-10-21
Prow cluster autoscaler stuck occasionally


lentzi90:
Due to this issue https://github.com/kubernetes/autoscaler/issues/6490#issuecomment-2076610573 the cluster autoscaler sometimes gets stuck in a loop where it thinks it doesn't have enough privileges to continue.
Deleting the pod gets it going again, but this depends on someone noticing it.

I propose that we either monitor the autoscaler every day to detect when it gets stuck OR we add the RBAC the controller thinks it needs until the upstream issue is fixed.

tuminoid:
Unless the RBAC it thinks it needs is very invasive, I think that is better workaround than constant manual monitoring. If implemented with RBAC, let's make sure we have revert PR or issue for reverting the change available right after merge.

/triage accepted

PR #903, opened 2024-11-18
Prow: Temporary autoscaler RBAC workaround

lentzi90:
There is a bug in the autoscaler that makes it get stuck on a permission
issue sometimes. It doesn't really need these permissions for our
use-case, but adding them should hopefully avoid it getting stuck.

See https://github.com/kubernetes/autoscaler/issues/6490

Fixes: #881

### DIFF
--- a/prow/cluster-resources/autoscaler.yaml
+++ b/prow/cluster-resources/autoscaler.yaml
@@ -175,3 +175,13 @@ rules:
   - list
   - update
   - watch
+# TODO (lentzi90): Remove this when the issue is solved
+# https://github.com/kubernetes/autoscaler/issues/6490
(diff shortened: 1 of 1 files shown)


tuminoid (commented):
/approve

Discussed offline and also seen this in the logs A LOT. Thanks for the workaround.

adilGhaffarDev:
thank you
/lgtm

lentzi90:
Thanks for the reviews!
Adding hold until I have the time to apply and check that it works
/hold

lentzi90:
It is now applied and looking good. I guess we will know for sure if it worked after a few days.
/hold cancel

PR #904, opened 2024-11-19
update basic-checks baseimage when building it

tuminoid:
Update baseimage when we build the basic-checks image. While we usually bump to quite new version, we also have repository_dispatch trigger in the workflow, so we can rebuild the image in which case baseimage is old and should get updated.

Quay is reporting a lot of vulnerabilities for this image already.


lentzi90 (commented):
/approve

Sunnatillo (commented):
/lgtm

Issue #871, opened 2024-10-01
Migrate from gcr.io to registry.k8s.io


tuminoid:
We're using gcr.io based Prow images in project-infra. k8s stopped publishing images 2 months ago, hence we're not getting autobumper to bump new images anymore.

Ref: https://github.com/metal3-io/project-infra/pull/869#issuecomment-2385001454

tuminoid:
It should be noted that Prow images do not have published production images available yet. Migration to registry.k8s.io is not done yet, and they look for volunteers to help.

tuminoid:
https://github.com/metal3-io/project-infra/pull/908 makes us use the staging registry as intemediate measure, as we haven't had any Prow image updates for 3 months now. There is activity in #prow community to complete the migration, but we can move to that when its done. We need some fixes from Prow sooner than that happens.

tuminoid:
/lifecycle frozen

Freezing this for now as it is waiting external actions.

tuminoid:
Updated title as per https://github.com/metal3-io/project-infra/issues/871#issuecomment-2493228783 we have moved to using Prow staging images, so the task is no longer to move from deprecated location, but from "unofficial" location. Community is still working in migration, so we'll wait.

PR #908, opened 2024-11-22
Prow: Switch to staging registry

lentzi90:
The GCR is about to be shut down. Prow already stopped publishing images there but have not transitioned to registry.k8s.io yet. Because of this we are switching to use the staging registry instead, which is the same that the k/k prow is also using.

Related to #871 (doesn't fix, since the issue is for migrating to registry.k8s.io)


lentzi90:
Changes are applied, except the utility images that will be auto-applied when this merges.

tuminoid (commented):
/approve

Sunnatillo (approved):
/lgtm

Perfect. Thank you

PR #910, opened 2024-11-22
Prow: Add Makefile to autobumper

lentzi90:
The autobumper is not checking Makefiles unless explicitly configured to do so. I have tested this with a dry-run locally.


tuminoid (commented):
/approve

Good catch!

prow/config/generic-autobumper-config.yaml
adilGhaffarDev, line 1: /lgtm

PR #912, opened 2024-11-25
Prow: Use GCR for commenter and label_sync images

lentzi90:
These components are not maintained in the new prow repo and have apparently not been pushed to the same staging registry. Instead they are pushed to the test-infra staging registry in GCR.

This commit changes the registry for the commenter and label_sync jobs and updates the autobumper to detect these images as well. I also had to bump the tag of the images since they have not pushed old image to the same registry.

### DIFF
--- a/prow/config/generic-autobumper-config.yaml
+++ b/prow/config/generic-autobumper-config.yaml
@@ -6,10 +6,14 @@ remoteName: "project-infra"
 includedConfigPaths:
 - "prow"
 targetVersion: "latest"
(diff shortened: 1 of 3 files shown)


tuminoid (commented):
/approve

lentzi90:
Label sync job applied and confirmed working. The periodic stale is handled as a ProwJob so it is autoapplied on merge.

smoshiur1237 (commented):
/lgtm

Issue #907, opened 2024-11-20
Prow: Migrate to CAPI operator


lentzi90:
We are currently relying in `clusterctl` imperative commands for managing CAPI/CAPO for prow. We should move to a declarative setup based on the operator instead. This is necessary for automating deployments in the future.

PR #917, opened 2024-11-26
Prow: Use CAPI Operator

lentzi90:
We have so far been relying on imperative commands to manage the lifecycle of CAPI/CAPO and cert-manager. This commit introduces the CAPI Operator with manifests to declaratively manage these components instead.

Fixes: #907

### DIFF
--- a/prow/README.md
+++ b/prow/README.md
@@ -77,8 +77,10 @@ You will need the following CLI tools in order to deploy and/or manage Prow:
 
 ### Folders and components
 
-There are four folders with kustomizations (`capo-cluster`, `cluster-resources`,
-`infra` and `manifests`). The `capo-cluster` folder contains everything needed for
+There are five folders with kustomizations (`capi`, `capo-cluster`,
+`cluster-resources`, `infra` and `manifests`). The `capi` folder contains a
+kustomization for cert-manager, the cluster API operator as well as the provider
+components we use. The `capo-cluster` folder contains everything needed for
 creating the Kubernetes cluster itself. In `cluster-resources`, you will find
 things the cluster needs to integrate with the cloud, i.e. the external
 cloud-provider for OpenStack and CSI plugin for Cinder. It also has the CNI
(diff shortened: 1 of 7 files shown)


lentzi90:
/hold
I want to make sure everyone is fine with the way this works before we merge.
Changes have not been applied yet.

tuminoid:
/cc @adilGhaffarDev @kashifest 
CAPI guys PTAL.

kashifest (commented):
awesome
/approve

prow/README.md
```diff
@@ -365,8 +367,10 @@ You may also have to create a keypair with the Metal3 CI ssh key.
 
    ```bash
    kind create cluster
-   clusterctl init --infrastructure=openstack:v0.11.1 --core=cluster-api:v1.8.5 \
-      --bootstrap=kubeadm:v1.8.5 --control-plane=kubeadm:v1.8.5
+   kubectl apply -k capi
+   # NOTE! You WILL need to apply multiple times before it is successful.
+   # This is because CRDs and webhooks must be in place before other
+   # resources can be added.
```
lentzi90, line 373: Note this specifically!
This is an opinionated way of working. I put all of it in one kustomization to keep things simple.
However, that kustomization can never be applied in one go. It will *always* fail the first time you apply, because you cannot create a Certificate or CoreProvider until the CRDs and webhooks are in place.
So the question is, would you rather have a script that does this or is it good like this? Or do you want something different?

To make it crystal clear, the way I would use this if I have to deploy prow from scratch is like this:

1. kind create cluster
2. kubectl apply -k capi <- This would fail
3. Wait a few seconds
4. kubectl apply -k capi <- This could potentially succeed but will probably fail
5. Wait a few seconds
6. kubectl apply -k capi <- This would most likely succeed
7. Continue to the next step

tuminoid: That is not something you do all the time, so documenting it clearly (like above comment) is enough IMO.

tuminoid (commented):
/lgtm

adilGhaffarDev:
LGTM. it looks great.

PR #921, opened 2024-11-27
add blunderbuss to Nordix repos

tuminoid:
Nordix support repos could use blunderbuss as well.

### DIFF
--- a/prow/config/plugins.yaml
+++ b/prow/config/plugins.yaml
@@ -10,6 +10,7 @@ plugins:
     plugins:
     - approve
     - assign
+    - blunderbuss
     - cat
     - dog
     - heart


prow/config/plugins.yaml
```diff
@@ -89,7 +90,7 @@ approve:
   lgtm_acts_as_approve: false
 
 blunderbuss:
-  max_request_count: 2
+  max_request_count: 3
```
lentzi90, line 93: I'm not quite sure this is a good idea. It will notify more people, which is not so nice. Also, the reason it may miss approvers is because we are supposed to first just get review, then when that is good ask for approval.

I agree with the issue though. Sometimes we get reviewers that are not active and the PR hangs because of that.
But I'm afraid this will make things worse because of "notification fatigue". If the busy reviewers get notified on everything they will eventually just ignore. Or it will take a long time to go through already reviewed PRs that they were notified on "just to make sure".

tuminoid: I can remove it for sure if you feel that way.

tuminoid:

Only Nordix addition is now in this repo, increasing the number of reviewers removed.

prow/config/plugins.yaml
lentzi90: Let's see what others think. I don't have that strong of an opinion :slightly_smiling_face:

tuminoid: I already dropped it. We can do it in another PR, if we feel like it. 

First and foremost I want blunderbuss on Nordix repos as they don't get any now and you don't remember to add any now when you expect blunderbuss to handle it :)

lentzi90 (commented):
/approve

Issue #922, opened 2024-11-28
Prow: stop using volumes and use local disk instead


tuminoid:
With the new private infrastructure servers deployed, Prow should stop using slow volumes and use local disk instead.

tuminoid:
@lentzi90 as discussed offline.

lentzi90:
Let's just confirm that we have capacity first. It would not be nice to reserve local disk for prow if that makes the jenkins tests starved for resources.

tuminoid:
For the reference, can you summarize how the disk space was spent right now? @lentzi90

lentzi90:
Sure! Prow is currently using [100 GiB](https://github.com/metal3-io/project-infra/blob/ae5d38f7cdd98bb460c73efdd166c7ce00ac43f4/prow/capo-cluster/openstackmachinetemplates.yaml#L70) volumes. The cluster consists of 1 control plane node and minimum 3 worker nodes at the moment. The workers can scale up to a total of 6. The bastion host does not use any volume.

**In total, that can be maximum 700 GiB. Most of the time we are at 400 GiB.**

(Outside of this is 1 volume for ghproxy, also 100 GiB. This is attached to a pod so using local disk for it may not be a good idea. It would lose all state every image bump or eviction.)

tuminoid:
Thank you. I think we are safe to do the change next week. Leave ghproxy with a volume.

PR #924, opened 2024-12-04
Prow: Use flavors with local disk

lentzi90:
To improve performance and reduce the risk of network issues due to volumes mounted over network, we switch to flavors with local disk. Root volumes are no longer needed.

Fixes: #922

### DIFF
--- a/prow/capo-cluster/infra-md.yaml
+++ b/prow/capo-cluster/infra-md.yaml
@@ -27,5 +27,5 @@ spec:
       infrastructureRef:
         apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
         kind: OpenStackMachineTemplate
-        name: prow-worker-v1-30-6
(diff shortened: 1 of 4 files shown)


lentzi90:

The changes are already applied. As you can see, the flavors are a bit bigger than what we used to have. If it turns out that we are wasting resources this way, we may ask for smaller flavors in the future. For now I think we can try this.

tuminoid (commented):
/approve

tuminoid:
> /override metal3-ubuntu-e2e-integration-test-main
> 
> The changes are already applied. As you can see, the flavors are a bit bigger than what we used to have. If it turns out that we are wasting resources this way, we may ask for smaller flavors in the future. For now I think we can try this.

Agree, let's start with this and get the prow monitoring done, then we have data to make decisions.

Rozzii (commented):
/lgtm

Issue #896, opened 2024-11-08
Prow: Monitoring


lentzi90:
It would be nice to have a monitoring solution for Prow. This would help us set more precise resource requests, send out alerts if there are issues and track metrics exposed by prow itself also. A good starting place is probably the monitoring solution they have in test-infra: https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/monitoring/README.md

At minimum this issue should provide a way to track prow job resource usage, normal container and node metrics. These metrics could then be used for alerting.

PR #928, opened 2024-12-13
Prow monitoring through kube-prometheus

lentzi90:
This adds the kube-prometheus stack to our prow cluster and configures it to monitor prowjobs on top of all that it does out of the box.

I have purposefully not squashed the commits to make it easier to follow the process but I can of course squash if that is preferred. The most interesting part is probably to compare the initial commit, that has the example.jsonnet unmodified, with the changes that comes later.

Another thing to note when reviewing: everything under `manifests` is *rendered*. I include it in the repository because it is what we apply in the cluster and because it is far easier to understand than the jsonnet code.  It also shows clearly what the changes to the jsonnet code does when comparing commits.

We should probably discuss if we want to keep the jsonnet code at all or if we just want to continue from the rendered manifests. We would then no longer be able to consume the upstream kubernetes-mixins and kube-prometheus, but we get one less tool and language to keep track of. It looks like [k8s.io has chosen this route](https://github.com/kubernetes/k8s.io/tree/98668f8f2003a19fb9fbd6d5bc1e4555a30d4086/infra/aws/terraform/prow-build-cluster/resources/monitoring).
In any case, I think we can merge it with the jsonnet first, before we decide how to handle it. It will be a good historic record at least that shows where it came from.

A couple of "controversial" or notable points:

- All in-cluster traffic is allowed on the OpenStack level. We did not need this before so it was disabled. Now prometheus needs to reach node-exporter on every node. We can limit further if necessary but I do believe this is a quite common policy.
- Grafana is configured with anonymous auth and everyone has the viewer role. This is the [same as k8s.io](https://github.com/kubernetes/k8s.io/blob/98668f8f2003a19fb9fbd6d5bc1e4555a30d4086/infra/aws/terraform/prow-build-cluster/resources/monitoring/grafana/config.yaml). The point is to make grafana practically read-only.
- Despite the above point, I have not yet exposed grafana. Should I?
- The prowjob dashboard is a bit outdated, but it is the same that is used in the k8s.io instances. We have a newer version of grafana though, that complains about it. Let's address this in the future.

Fixes: #896


tuminoid:
> /cc @tuminoid

Thanks @lentzi90, I'll take a look. 99 files changed, might take a while :) 

> This adds the kube-prometheus stack to our prow cluster and configures it to monitor prowjobs on top of all that it does out of the box.
> 
> I have purposefully not squashed the commits to make it easier to follow the process but I can of course squash if that is preferred. The most interesting part is probably to compare the initial commit, that has the example.jsonnet unmodified, with the changes that comes later.

This is certainly a case where we don't want squashed commits. It is indeed very helpful to have them in the way you committed them.

> Another thing to note when reviewing: everything under `manifests` is _rendered_. I include it in the repository because it is what we apply in the cluster and because it is far easier to understand than the jsonnet code. It also shows clearly what the changes to the jsonnet code does when comparing commits.
> 
> We should probably discuss if we want to keep the jsonnet code at all or if we just want to continue from the rendered manifests. We would then no longer be able to consume the upstream kubernetes-mixins and kube-prometheus, but we get one less tool and language to keep track of. It looks like [k8s.io has chosen this route](https://github.com/kubernetes/k8s.io/tree/98668f8f2003a19fb9fbd6d5bc1e4555a30d4086/infra/aws/terraform/prow-build-cluster/resources/monitoring). In any case, I think we can merge it with the jsonnet first, before we decide how to handle it. It will be a good historic record at least that shows where it came from.

I agree, we need the historic portion at least. Initially, I dislike completely detaching, but I'll comment further on this after reading the actual code. I can see the point of going with the rendered tho,

> A couple of "controversial" or notable points:
> 
> * All in-cluster traffic is allowed on the OpenStack level. We did not need this before so it was disabled. Now prometheus needs to reach node-exporter on every node. We can limit further if necessary but I do believe this is a quite common policy.

I think we can add some NetworkPolicies to enforce this as a simple hardening?

> * Grafana is configured with anonymous auth and everyone has the viewer role. This is the [same as k8s.io](https://github.com/kubernetes/k8s.io/blob/98668f8f2003a19fb9fbd6d5bc1e4555a30d4086/infra/aws/terraform/prow-build-cluster/resources/monitoring/grafana/config.yaml). The point is to make grafana practically read-only.

This is good approach. We're open source, our Prow is generally accessible, what would hiding it gain really? We have the ability to configure it on the backend side anyways.

> * Despite the above point, I have not yet exposed grafana. Should I?

Maybe we can have that in separate PR after this merges, to make it separate decision and self-documentary if it in future needs to be removed?

tuminoid:
> Someone could DDoS it I guess, but they could already do that with Prow so 🤷 Loading dashboards in grafana also queries prometheus, so there is a certain load and some risk that this could be used to gain deeper access in the cluster I guess. Generally this does not seem to be a big concern though, since there are loads of public grafana instances out there.

Indeed, our tiny Prow can be DoS'd, no need to DDoS even. Grafana and Prometheus are not abandonware, so if we keep them up to date, and do some basic hardening on the config, we should be fine.

tuminoid (commented):
I've forgotten what was the status of applying this... @lentzi90 was this ready to go, applied already? 

/approve
/hold

lentzi90:
Yes it is already in the cluster :slightly_smiling_face: 
/hold cancel

PR #931, opened 2024-12-17
add IPAM 1.9 release to Prow

tuminoid:
Add IPAM 1.9 release to Prow.


tuminoid:
/hold
Needs some JJB first.

tuminoid:
/unhold
This also defines basic jobs, and we're merging stuff there now, so this needs to go in, even without e2e being ready.
/cc @kashifest

PR #937, opened 2024-12-18
:bug: Fix error on reusable container image build workflow

mquhuy:
This workflow is giving incorrect tags, since a trailing / is left at the REF_NAME, which turns into trailing _ in the tags.


tuminoid (commented):
/approve

tuminoid (commented):
/lgtm

PR #944, opened 2025-01-09
:seedling:  Add new reusable release note generator

mquhuy:
This tool uses `go-github` package to fetch PR information from upstream using github API, hence doesn't need to be triggered from inside a repo, and can work from anywhere. It also avoids the usage of `os.Exec()`, and rely on go-native methods instead.

Compare to the current script that we are having in all releasing repositories, this tool:
- Move from using local repo to accessing remote repo via API.
- Is simpler: multiple custom functionalities are rewritten in simpler and more native way.
- Take custom repo name and repo owner: It works with multiple repos, and can work on a fork if necessary.
- Doesn't use os.Exec(): less error-prone and doesn't need to run inside the repo.
- Doesn't assume local repo configuration, like uptream `git fetch` before running.
- Doesn't require GITHUB_TOKEN for public repo. Still support GITHUB_TOKEN if we're to run this tool multiple times in a row.
- Is more reliable as everything is fetched from github API.
- As it works with custom repo name, we no longer need to copy the script to all of our repositories.

### DIFF
--- a/hack/release/main.go
+++ b/hack/release/main.go
@@ -1,0 +1,216 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/blang/semver"
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+/*
+This tool prints all the titles of all PRs from previous release to HEAD.
+This needs to be run *before* a tag is created.
+
+Use these as the base of your release notes.
+*/
+
+const (
+	features        = ":sparkles: New Features"
+	bugs            = ":bug: Bug Fixes"
+	documentation   = ":book: Documentation"
+	warning         = ":warning: Breaking Changes"
+	other           = ":seedling: Others"
+	unknown         = ":question: Sort these by hand"
+	superseded      = ":recycle: Superseded or Reverted"
+	warningTemplate = ":rotating_light: This is a %s. Use it only for testing purposes.\nIf you find any bugs, file an [issue](https://github.com/%s/%s/issues/new/).\n\n"
+)
+
+var (
+	outputOrder = []string{
+		warning,
+		features,
+		bugs,
+		documentation,
+		other,
+		unknown,
+		superseded,
+	}
+	releaseTag     string
+	repoOwner      string
+	repoName       string
+	semVersion     semver.Version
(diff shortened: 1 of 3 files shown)


hack/release/main.go
```diff
@@ -0,0 +1,217 @@
...
+
+const (
+	features        = ":sparkles: New Features"
+	bugs            = ":bug: Bug Fixes"
+	documentation   = ":book: Documentation"
+	warning         = ":warning: Breaking Changes"
+	other           = ":seedling: Others"
+	unknown         = ":question: Sort these by hand"
+	superseded      = ":recycle: Superseded or Reverted"
+	warningTemplate = ":rotating_light: This is a %s. Use it only for testing purposes.\nIf you find any bugs, file an [issue](https://github.com/metal3-io/cluster-api-provider-metal3/issues/new/).\n\n"
```
adilGhaffarDev, line 49: link to issue creation should be different for different repos. Let's not make it constant here.

hack/release/main.go
```diff
@@ -0,0 +1,217 @@
...
+			fmt.Println()
+		}
+	}
+
+	// close the details tag if we had it open
+	if len(semVersion.Pre) > 0 {
+		fmt.Printf("</details>\n\n")
+	}
+
+	fmt.Printf("The container image for this release is: %v\n", releaseTag)
```
tuminoid, line 215: Capm3 has also mariadb image tag here, maybe add an if about it?

Adding it manually is error-prone.

hack/release/main.go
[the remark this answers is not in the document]
mquhuy: Oops. It seems the old code used spaces, but my editor uses tabs, and the two looked identical in my setup. Fixed it now.

hack/release/main.go
```diff
@@ -0,0 +1,219 @@
...
+    // Append commit message and PR number
+    lines := strings.Split(commit.GetCommit().GetMessage(), "\n")
+    body := lines[len(lines)-1]
+    if body == "" {
+      continue
+    }
+    var key string
+    switch {
+    case strings.HasPrefix(body, ":sparkles:"), strings.HasPrefix(body, "✨"):
+      key = features
```
tuminoid, line 156: Indents are wonky again.

mquhuy: Thanks. I fixed it again.

mquhuy:
> Can you update the description that describes how it actually works and what is the difference to the old way? The current bullet list really doesn't convey the design change.

Yes. Could you check if the current description makes more senses?

tuminoid:
> > Can you update the description that describes how it actually works and what is the difference to the old way? The current bullet list really doesn't convey the design change.
> 
> Yes. Could you check if the current description makes more senses?

Yes, it a bit better. It would move from reading a local repo to accessing the repo over API.

 How would this be used? Do we make released project's makefile depend on project-infra when we run "make release-notes", or ... ?

mquhuy:
> Yes, it a bit better. It would move from reading a local repo to accessing the repo over API.
 
Thank you, I will update that.

> How would this be used? Do we make released project's makefile depend on project-infra when we run "make release-notes", or ... ?

In the make file, we can do something like `go install <project-infra>/hack/release@main`, then use the binary to run `make release-notes`.

tuminoid:
If this gets merged, https://github.com/metal3-io/project-infra/issues/951 should be done after.

tuminoid (changes requested):
This still needs review and probably a new owner, blocking merging for now.

tuminoid:
When this work is continued by someone, you need to consider https://github.com/metal3-io/ironic-image/issues/696 and other repo specific improvements to the generator script.

PR #945, opened 2025-01-09
Prow: Use latest CRDs

lentzi90:
The CRDs were pinned because of an upstream issue. It has been solved, so we can go back to using the latest.
The upstream issue is documented here: https://github.com/kubernetes-sigs/prow/issues/181

I have updated the CRDs in the cluster also.
For the future we should also come up with a system for keeping them up to date. For example, update the CRDs every time we bump the images.


tuminoid (commented):
/lgtm

PR #953, opened 2025-01-15
:seedling: Add scalability test prow triggers

mquhuy:
This commit adds the new scalability tests to prow

### DIFF
--- a/prow/config/jobs/metal3-io/cluster-api-provider-metal3-release-1.9.yaml
+++ b/prow/config/jobs/metal3-io/cluster-api-provider-metal3-release-1.9.yaml
@@ -200,3 +200,9 @@ presubmits:
     agent: jenkins
     always_run: false
     optional: true
+  - name: metal3-centos-e2e-feature-test-release-1-9-scalability
+    branches:
+    - release-1.9
+    agent: jenkins
+    always_run: false
+    optional: true
--- a/prow/config/jobs/metal3-io/cluster-api-provider-metal3.yaml
+++ b/prow/config/jobs/metal3-io/cluster-api-provider-metal3.yaml
@@ -191,6 +191,12 @@ presubmits:
     agent: jenkins
     always_run: false
     optional: true
+  - name: metal3-centos-e2e-feature-test-main-scalability
+    branches:
+    - main
+    agent: jenkins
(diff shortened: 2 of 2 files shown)


lentzi90 (commented):
/approve

kashifest (commented):
/lgtm

prow/config/jobs/metal3-io/cluster-api-provider-metal3-release-1.9.yaml
```diff
@@ -200,3 +200,10 @@ presubmits:
     agent: jenkins
     always_run: false
     optional: true
+  # name: metal3-centos-e2e-feature-test-release-1-9-scalability
```
tuminoid, line 203: These comments do nothing :)

mquhuy: Imo they're still useful as separation.

tuminoid: In release branch config they seem removed altogether.

They're supposed to show the template format, like they do in main config.

tuminoid: /hold

mquhuy: That's not something we have agreed on. We started adding those comments back when the prow config grows so big and unorganized, so the point of having them is just so that we can tell apart which from which, we didn't discussed the "rules" of where they should be and how they should look like/

Since you insist I'll remove these comments now, but IMO adding the comments here really help making the separation easier to follow.

tuminoid (commented):
/lgtm

prow/config/jobs/metal3-io/cluster-api-provider-metal3-release-1.9.yaml
tuminoid: Convention is form of agreement as well.

Zero information comments are worth nothing, they're actually worth negative since they need to be maintained not to be outdated.

If you want to use separator, use empty line.

PR #955, opened 2025-01-22
add IRSO to milestoneapplier plugin config

tuminoid:
IRSO was missing from milestone applier config. Add it there, which makes any PR landing in main be part of the next release, which is v0.2.


lentzi90 (commented):
/approve

PR #960, opened 2025-02-03
Enables optionally multiplatform container builds

e-minguez:
https://github.com/metal3-io/ironic-image/issues/615

### DIFF
--- a/.github/workflows/container-image-build.yml
+++ b/.github/workflows/container-image-build.yml
@@ -27,6 +27,16 @@ on:
         description: "Whether to push the image afterwards"
         type: boolean
         default: true
+      multiPlatform:
+        required: false
+        description: "Enable multiPlatform builds for the container image"
+        type: boolean
+        default: false
+      platform:
+        required: false
+        description: "Architecture platforms for the image build (linux/amd64,linux/arm64, etc.)"
+        type: string
+        default: ""
     secrets:
       QUAY_USERNAME:
         required: true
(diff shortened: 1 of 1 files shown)


.github/workflows/container-image-build.yml
```diff
@@ -27,6 +27,16 @@ on:
...
+      multiPlatform:
+        required: false
+        description: "Enable multiPlatform builds for the container image"
+        type: boolean
+        default: false
+      platform:
+        required: false
+        description: "Architecture platforms for the image build (linux/amd64,linux/arm64, etc.)"
+        type: string
+        default: ""
```
kashifest, line 39: I am thinking if we would need a default value here

kashifest:
@e-minguez thanks a lot for the PR, please check the DCO failure and fix it

.github/workflows/container-image-build.yml
e-minguez: I _guess_ (honestly, I didn't tested it) that if you don't set `multiPlatform: false` this will be ignored and built against the arch where it has been executed (`linux/amd64` by default IIRC)

e-minguez: I _guess_ (honestly, I didn't tested it) that if you don't explicitly set `multiPlatform: true` this will be ignored and built against the arch where it has been executed (`linux/amd64` by default IIRC)

lentzi90:
The prow-config test failure is irrelevant. We have some issue with the infra. Please ignore for now

Rozzii (commented):
/lgtm
Why not...
I am not sure this will actually produce a working container  but let's see.

tuminoid (commented):
/hold

I need some time to look at this. I'm not fond of possibly breaking all builds.

lentzi90 (commented):
It has been a few months. I think we should take this in and be prepared to revert in case there are issues.
/approve

PR #962, opened 2025-02-05
Use server 04 instead of 05 in BML test

kashifest:
Switching to 04 from 05 since server 05 has started to show some hardware issues


tuminoid (commented):
/approve

kashifest:
I think we are good with 05 now after the fixes. Lets close this for now
/close

PR #963, opened 2025-02-07
Bump k8s to v1.32.1 and CRIO to v1.32.1

kashifest:
Bumping k8s and CRIO to latest patch version for v1.32.x


lentzi90 (commented):
/approve

PR #967, opened 2025-03-04
Prow: Add tolerations and nodeSelectors for CAPI related components and bump versions

lentzi90:
Make sure that all CAPI related components run on the infra nodes by adding tolerations and nodeSelectors to them.
Bump cert-manager and CAPI operator versions and update README to always use the operator way of deploying CAPI providers.
Bump CAPI core, KBT and KCP to v1.9.5 and CAPO to v0.12.1.
CAPO has a new external dependency, ORC, so this is also added, plus tolerations and nodeSelector for it.

Changes are already applied in the cluster.

### DIFF
--- a/prow/README.md
+++ b/prow/README.md
@@ -422,8 +422,10 @@ You may also have to create a keypair with the Metal3 CI ssh key.
 1. Make cluster self-hosted
 
    ```bash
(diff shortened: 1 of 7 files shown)


tuminoid (commented):
/approve

PR #968, opened 2025-03-05
Prow: Upgrade kubernetes, ingress, calico and more

lentzi90:
See individual commit messages for more details.
This PR does the following:
- Upgrade ingress-nginx to v1.12.0
- Upgrade Calico
- Upgrade cloud-provider integrations and cluster-autoscaler
- Upgrade kubernetes to v1.31.6
- Upgrade nodes and bastion to Ubuntu 24.04

The changes are already applied.

### DIFF
--- a/prow/infra/add-args.yaml
+++ b/prow/infra/add-args.yaml
@@ -1,4 +1,0 @@
-# See https://github.com/kubernetes/ingress-nginx/issues/10572
-- op: add
-  path: /spec/template/spec/containers/0/args/-
-  value: --enable-annotation-validation=true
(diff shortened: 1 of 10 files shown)


prow/infra/add-args.yaml
lentzi90, line 1: This is no longer needed since the upstream has changed the defaults. See changelog here: https://github.com/kubernetes/ingress-nginx/releases/tag/controller-v1.12.0

tuminoid (commented):
/approve

Looks clean to me, and good that we are using local-disk enabled flavors too.

PR #971, opened 2025-03-12
prow-test-dumper: handle completely missing branch protection case

tuminoid:
In case a branch is not having any branch protection (it is not defined at all, or matching any branch protection wildcards), we just exited before. Now, it will return empty set, so we correctly get notified that tests are missing, and also the processing continues to further jobs defined.


lentzi90 (commented):
/approve

kashifest (commented):
/lgtm

PR #972, opened 2025-03-14
Add buildArgs to docker build conditionally

elfosardo:
Adds the possibiltiy to specify and env variable IMAGE_BUILD_ARGS to be used as list of buildArg when building the containers.


.github/workflows/container-image-build.yml
```diff
@@ -81,7 +81,22 @@ jobs:
...
 
+    - name: Build ${{ inputs.image-name }} image with buildArgs
+      if: ${{ env.IMAGE_BUILD_ARGS }}
+      uses: mr-smithers-excellent/docker-build-push@59523c638baec979a74fea99caa9e29d97e5963c # v6.4
+      with:
+        image: metal3-io/${{ inputs.image-name }}
+        tags: ${{ env.IMAGE_TAGS }}
+        directory: ${{ inputs.dockerfile-directory }}
+        dockerfile: ${{ inputs.dockerfile-directory }}/Dockerfile
+        buildArgs: ${{ env.IMAGE_BUILD_ARGS }}
```
tuminoid, line 92: I wonder if it would just work even if the env doesn't exist, and we don't need to do anything here, but add this one line to existing step?

elfosardo: the problem I see it's how that is parsed in https://github.com/mr-smithers-excellent/docker-build-push/blob/master/src/docker.js#L64
even in case env.IMAGE_BUILD_ARGS is not defined but still the job does not complain about that, --build-arg requires an argument otherwise it fails, so one possible workaround would be having a default empty string (''), which IMHO would produced a bit confusing output as we'll have `--build-arg ''`
if you think it's a better alternative I can change the PR to use that

tuminoid:
```javascript
  if (buildOpts.buildArgs) {
    const argsSuffix = buildOpts.buildArgs.map(arg => `--build-arg ${arg}`).join(' ');
    buildCommandPrefix = `${buildCommandPrefix} ${argsSuffix}`;
  }
```

In Javascript empty string `''` would evaluate as false, so it would not add a phony `--build-arg` there. Hence I think we can have empty string as default, and not duplicate the code here.

tuminoid (commented):
/approve

This should work. I'm interested how you cleanly inject the baseimage and sha to the env on the caller side though.

tuminoid:
> This should work. I'm interested how you cleanly inject the baseimage and sha to the env on the caller side though.

Never mind, centos images are not pinned so we don't need to maintain the pinning in the source workflow.

Rozzii (commented):
/lgtm

PR #981, opened 2025-04-02
Remove unnecessary git credentials

lentzi90:
We have been using username/password for checking out all repos. This is not necessary for public repos and doesn't seem to make any difference. The password has been changed without impacting any of the jobs... so this commit removes the use of these credentials.


tuminoid (commented):
If we somehow face rate-limiting due anonymous cloning, then we need to reimplement this, but as long as we don't (and we shouldn't be even close), this is 
/approve

PR #982, opened 2025-04-03
Bump openjdk to 17 in CI images

lentzi90:
Looks like jenkins is running JVM 17 while we are still on 11. Bumping to stay compatible.


tuminoid (commented):
/approve

17 matches Jenkins version used, so let's go for that.

PR #983, opened 2025-04-04
Rotate authorized ssh key in images

lentzi90:
Time to change the authorized key we bake in the images.
Side note: I think we should try to get away from this habit of baking it in the image. We can inject it as needed with cloud-init instead.


tuminoid (commented):
/approve

kashifest (commented):
/lgtm

PR #989, opened 2025-04-23
Fix image build refspec for PR jobs

lentzi90:
When triggering jobs on PRs, we need to include the PR commit(s) in the refspec. Otherwise we cannot checkout the relevant code.


lentzi90:
It is failing when checking out, as seen [here](https://jenkins.nordix.org/view/Metal3/job/metal3-ci-image-building/1/). I have now added the same refspec that we use here: https://github.com/metal3-io/project-infra/blob/877163671f3500934db4e2ad7989c630930c5867/jenkins/jobs/dev_env_integration_tests.pipeline#L12-L15

/cc @tuminoid

tuminoid (commented):
Makes sense. Just commented about this on the other PR.

/approve

PR #984, opened 2025-04-04
Bump ubuntu images to 24.04

lentzi90:
We have already long supported ubuntu 24.04. Time to bump all the images also.

Fixes: #954

PR #990, opened 2025-04-23
Fix image building refspec and upload

lentzi90:
When triggering jobs on PRs, we need to include the PR commit(s) in the
refspec. Otherwise we cannot checkout the relevant code. Pre-build-merge
is also removed to keep the behavior similar to other jobs.
This commit also avoids uploading images from PR tests and making them "latest".
It also fixes the labels for the CI image verification step so that the
label is separate for ubuntu and centos images.

Last but not least, this tweaks the logic for image retention.
CI images are uploaded as metal3-ci-<os>-staging for verification.
In periodic jobs, the staging images are then renamed to *-latest after
successful verification. This is not done in PR jobs.

Any existing *-staging images are deleted before uploading a new.
Otherwise we risk mixing them up with the previously used CI images and
fill the retention queue with staging images instead of the real
"production" images because the logic of the upload script is to rename
them back to their "original" name when there are conflicts.

See https://github.com/metal3-io/project-infra/pull/984#issuecomment-2825135017

### DIFF
--- a/jenkins/jobs/image_building.pipeline
+++ b/jenkins/jobs/image_building.pipeline
@@ -1,9 +1,17 @@
 script {
   ci_git_branch = (env.PULL_PULL_SHA) ?: "main"
+  ci_git_base = (env.PULL_BASE_REF) ?: "main"
   ci_git_url = "https://github.com/metal3-io/project-infra.git"
+  refspec = '+refs/heads/' + ci_git_base + ':refs/remotes/origin/' + ci_git_base + ' ' + ci_git_branch
 }
 
 pipeline {
+  options {
+    // We verify the CI images by uploading and testing them as jenkins agents.
+    // If multiple jobs would run in parallel, they may overwrite each others image
+    // and/or delete them before the test is finished.
+    disableConcurrentBuilds()
+  }
   agent none
   environment {
     IMAGE_TYPE = "${env.IMAGE_TYPE}"
@@ -50,11 +58,10 @@ pipeline {
                 extensions: [
                   [$class: 'WipeWorkspace'],
                   [$class: 'CleanCheckout'],
-                  [$class: 'CleanBeforeCheckout'],
-                  [$class: 'PreBuildMerge', options: [mergeTarget: 'main']]
+                  [$class: 'CleanBeforeCheckout']
                 ],
                 submoduleCfg: [],
-                userRemoteConfigs: [[url: ci_git_url]]
+                userRemoteConfigs: [[url: ci_git_url, refspec: refspec]]
               ])
             }
           }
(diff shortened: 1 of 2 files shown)


lentzi90:
Why didn't I think of triggering the build on the last PR :facepalm: 
/test metal3-ci-image-building

lentzi90:
Figured out a working config using the Replay functionality in jenkins. Successful clone can be seen in https://jenkins.nordix.org/view/Metal3/job/metal3-ci-image-building/6/

lentzi90:
/hold
It failed on uploading the image. Looks like there is some issue with the name because it could not find the image after uploading. Checking

lentzi90:
This will again fail because the checkout is from main (which doesn't have the fix). I'll do a replay again and see if it goes better with the new adjustments
/test metal3-ci-image-building

lentzi90:
https://gerrit.nordix.org/c/infra/cicd/+/23482 is needed for this to work

tuminoid:
> https://gerrit.nordix.org/c/infra/cicd/+/23482 is needed for this to work

+2, @kashifest to take second look and submit if ok.

jenkins/jobs/image_building.pipeline
```diff
@@ -71,47 +72,53 @@ pipeline {
...
-              timeout(time: 3, unit: 'HOURS')
+              timeout(time: 30, unit: 'MINUTES')
             }
             when {
-              expression { env.IMAGE_TYPE == 'node' }
+              expression { env.IMAGE_TYPE == 'ci' }
             }
             steps {
-              echo "Testing new ${IMAGE_OS} node image"
+              // NOTE: We delete any existing *-staging images before uploading the new one.
```
tuminoid, line 83: Can we have a situation where we run staging tests from multiple PR and these (try) delete images being used? I don't see us limiting this to 1 at a time?

tuminoid (commented):
Very good improvements. Added question.

jenkins/jobs/image_building.pipeline
lentzi90: That is a good point. Let me check how to limit it to one, like we do for BML

lentzi90:
The test will unfortunately keep failing here because of the pre-build-merge that is still enabled on main (removed in this PR). I will do a replay in Jenkins with the code from the PR to test it

jenkins/jobs/image_building.pipeline
lentzi90: Fixed

lentzi90:
/test metal3-ci-image-building
Still expecting failure... will do a replay that will hopefully work

lentzi90:
Alright I think we are good to go here.
You can see the replay run where at least the centos image passed here: https://jenkins.nordix.org/blue/organizations/jenkins/metal3-ci-image-building/detail/metal3-ci-image-building/26/pipeline/147
(Ubuntu seem to be stuck building a container image but I think it has gone far enough to show that the pipeline at least is working)
/hold cancel

tuminoid (commented):
Good job fixing it, and I see concurrent builds is also disabled now.

/lgtm

PR #995, opened 2025-04-25
:seedling:  Uplift Kubernetes version to v1.33.0 for image building

peppi-lotta:
Uplift Kubernetes version to v1.33.0 for image building


lentzi90:
The good news is that the test passed and the pipeline logic worked well (it did not upload the image because this was a PR job).
The bad news is that the node image verification is not checking what we want it to check.
```
Testing CENTOS_9_NODE_IMAGE_K8S_v1.32.1
```
The kubernetes version is coming from jjb. I think we will need to change it so that the verification step unconditionally tests the image that was built.

Anyway, that has nothing to do with this PR. I will keep working on it separately.
/approve

tuminoid (commented):
/lgtm

PR #996, opened 2025-04-29
Don't clean the yum cache while building CentOS node image

kashifest:
Had the add the clean cache steps since the repo was not found previously but now it is taking too much time to update the repos after the cleaned cache. Lets see if this is at all needed now.


tuminoid (commented):
/approve

lentzi90 (commented):
/lgtm

PR #998, opened 2025-04-29
:seedling: Add IPAM release-1.10 to Prow

peppi-lotta:
Add IPAM release-1.10 to Prow

### DIFF
--- a/prow/config/jobs/metal3-io/ip-address-manager-release-1.10.yaml
+++ b/prow/config/jobs/metal3-io/ip-address-manager-release-1.10.yaml
@@ -1,0 +1,186 @@
+presubmits:
+  metal3-io/ip-address-manager:
+  - name: gomod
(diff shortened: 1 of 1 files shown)


prow/config/jobs/metal3-io/ip-address-manager-release1.10.yaml
tuminoid, line 1: This is copied from release-1.9, but it should be copied from main.

tuminoid, line 1: Also the filename is incorrect, missing - between release and 1.10

tuminoid: Filename is still incorrect.

lentzi90 (commented):
/approve

PR #1018, opened 2025-05-21
add org.opencontainers.image.version label

tuminoid:
Add org.opencontainers.image.version label to all containers built via the container-image-build.yml workflow. All other labels are already set via Dockerfile in repo in question, but version should be kept up to date, so it is better to have it dynamically added at build time.


lentzi90 (commented):
/approve

PR #1019, opened 2025-05-26
Bump CAPM3 main branch prow jobs to use go v1.24

kashifest:
Bumping the same also in CAPM3.


tuminoid (commented):
We'd probably want to  do the same in release branches while at it?

Golang goes EOL in all of them the same.

tuminoid (commented):
/approve

PR #1020, opened 2025-05-26
bump basic-checks image to golang:1.24.3

tuminoid:
Bump basic-checks image to golang:1.24.3 to support bumping golang in BMO later. This new image is not used by anyone or changing anything, until BMO and its prow jobs are bumped to Go 1.24. This is just pre-building the image, so it doesn't explode when we do.


Rozzii (commented):
/lgtm

PR #1021, opened 2025-05-26
fix container tags for main

tuminoid:
Since merging the labeling PR, it also moved the defaults tags in a way for main that they are empty. Fix that.


kashifest:
not sure if I understand the changes here, would wait for @lentzi90 or  @adilGhaffarDev to confirm

tuminoid:
> not sure if I understand the changes here, would wait for @lentzi90 or @adilGhaffarDev to confirm

L95 has empty ${BASE_TAG} if only defined in L98. Just move the variable up, so it is populated early enough.

PR #1025, opened 2025-05-27
Bump BMO prow jobs to use go v1.24

kashifest:
Bumps golang to v1.24 for prow jobs for following branches

- main
- release 0.10
- release 0.9


tuminoid (commented):
/approve

PR #1028, opened 2025-05-28
Use nordix artifactory proxy to download IPA

kashifest:
Lately, we have seen a lot of delay, flakes and failures while downloading IPA from opendev upstream. Now we have setup a proxy in Nordix rtfactory to proxy the IPA from upstream and use it in our tests in a faster way when cached.


smoshiur1237 (approved):
/lgtm

Rozzii:
Let's see what breaks after this. :D 
/hold cancel

Rozzii (commented):
/lgtm

PR #1042, opened 2025-06-17
add support for ironic-image release-30.0 branch

tuminoid:
Add CI support for ironic-image release-30.0 branch. It will be tested versus CAPM3 main, until CAPM3 v1.11 minor is out, and then changed to be tested versus that.


lentzi90 (commented):
/approve

PR #1054, opened 2025-07-02
Add RootDeviceHint on bmhs in BML tests

Sunnatillo:
When we changed RAID config to 0 in bml it started causing an issue. Sometimes After provisioning server cannot find OS in the disk. After we realized boot checks only first disk, and sometimes disk is written to second disk. To solve the issue we are adding rootDeviceHint to bmhs to make sure OS will always be written to first disk.

PR also removes server 2 info as it used as provisioning host


Rozzii (commented):
/lgtm

PR #1060, opened 2025-07-31
Add pipeline for testing with k8s' pre-release versions

erjavaskivuori:
Add new pipeline for testing metal3 with pre-release Kubernetes versions. In the pipeline

- new node image is built with the Kubernetes version that is tested
- the node image is not uploaded, and instead the local image is used when running CAPM3 e2e tests

Edited the image building installations to allow using pre-release versions.

### DIFF
--- a/jenkins/image_building/build-image.sh
+++ b/jenkins/image_building/build-image.sh
@@ -16,15 +16,33 @@ trap cleanup EXIT
 # Make sure we run everything in the repo root
 cd "${REPO_ROOT}" || true
 
+export AGENT_OS="${AGENT_OS:-"ubuntu"}"
+
 export IMAGE_OS="${IMAGE_OS}"
 export IMAGE_TYPE="${IMAGE_TYPE}"
 
-# Disable needrestart interactive mode
-sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null || true
+# Jenkins agent OS-specific package installation and configuration
+if [[ "${AGENT_OS}" == "ubuntu" ]]; then
+  # Disable needrestart interactive mode
+  sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null || true
 
-sudo apt-get update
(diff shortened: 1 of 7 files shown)


lentzi90:
This is touching some files that are related to the normal node-image-building test so let's run that test also
/test metal3-node-image-building

lentzi90 (commented):
/approve
Adding hold to make sure the test has time to complete
/hold

tuminoid:
```
+ sudo yum install -y epel-release
sudo: yum: command not found
```
Something is fundamentally broken with the node build?

jenkins/image_building/build-image.sh
```diff
@@ -19,12 +19,28 @@ cd "${REPO_ROOT}" || true
...
+  # Disable needrestart interactive mode
+  sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null || true
+
+  sudo apt-get update
+  sudo apt-get install -y python3-dev python3-pip python3-venv qemu qemu-kvm
+
+  python3 -m venv venv
+elif [[ "${IMAGE_OS}" == "centos" ]]; then
+  # Install EPEL repository for additional packages
+  sudo yum install -y epel-release
```
tuminoid, line 33: See the error about this.

lentzi90:
Yeah something strange is going on here. I don't understand it. It is like it is using the wrong agent/image, but this PR does not touch that

jenkins/image_building/build-image.sh
```diff
@@ -19,12 +19,28 @@ cd "${REPO_ROOT}" || true
 export IMAGE_OS="${IMAGE_OS}"
 export IMAGE_TYPE="${IMAGE_TYPE}"
 
-# Disable needrestart interactive mode
-sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null || true
+# OS-specific package installation and configuration
```
tuminoid, line 22: This is run on agent defined always to be ubuntu, IMAGE_OS should not be if'd here?

erjavaskivuori: I need to make the appropriate changes here to make the "normal" image building pipeline to work with these changes. On the normal pipeline we would always use ubuntu but on this test we would do the image building also on centOS. See my answer on other comment for longer explanation.

PR #1067, opened 2025-08-18
Add prow keywords for k8s pre-release testing

erjavaskivuori:
Add prow keywords for k8s pre-release testing for BMO, IPAM and CAPM3 and update the naming for project-infra.

PR #1069, opened 2025-08-20
Automatically detect agent's OS

erjavaskivuori:
Remove AGENT_OS variable from pipelines and add detection to the image building script.

### DIFF
--- a/jenkins/image_building/build-image.sh
+++ b/jenkins/image_building/build-image.sh
@@ -16,7 +16,11 @@ trap cleanup EXIT
 # Make sure we run everything in the repo root
 cd "${REPO_ROOT}" || true
 
-export AGENT_OS="${AGENT_OS:-"ubuntu"}"
+if [ -r /etc/os-release ]; then
+  # shellcheck disable=SC1091
+  . /etc/os-release
+  AGENT_OS="${ID}"
+fi
 
 export IMAGE_OS="${IMAGE_OS}"
 export IMAGE_TYPE="${IMAGE_TYPE}"
--- a/jenkins/jobs/image_building.pipeline
+++ b/jenkins/jobs/image_building.pipeline
@@ -73,7 +73,6 @@ pipeline {
               echo "Building ${IMAGE_OS} ${IMAGE_TYPE} image"
               script {
                 sh """
-                def AGENT_OS = "ubuntu"
                 ./jenkins/image_building/build-image.sh
                 """
               }
--- a/jenkins/jobs/pre_release_tests.pipeline
+++ b/jenkins/jobs/pre_release_tests.pipeline
@@ -18,7 +18,6 @@ script {
 pipeline {
   agent { label "metal3ci-8c16gb-${IMAGE_OS}" }
   environment {
-    AGENT_OS = "${IMAGE_OS}"
     REPO_ORG = "${env.REPO_OWNER}"
     REPO_NAME = "${env.REPO_NAME}"
     UPDATED_REPO = "${UPDATED_REPO}"


tuminoid:
> Just a note that the pre-release pipeline doesn't work with the keyword I used and can be only tested after #1067 is merged.
> 
> However, I tested the node and CI image building now and it seems to work with this change. They are still running but have passed the phases where the changes from this PR would affect the run:
> 
> * replay of node image building with pipeline changes: https://jenkins.nordix.org/view/Metal3/job/metal3-node-image-building/63/
> * replay of CI image building with pipeline changes: https://jenkins.nordix.org/view/Metal3/job/metal3-ci-image-building/42/
> 
> but once again centos is having some issues that are not related to the changes:
> 
> ```
> qemu-img: error while reading at byte 6186532864: Input/output error
> ```
> 
> This exactly same error occurs on node and CI image building. I have seen same errors before and without doing anything they went away at some point so not sure if there is anything we can do 😕

Since this is somehow related to the image used,  can it simply be that the image where our names point to do not really work, and there is some sort of image length mismatch that goes undetected and it just crashes?

This is now going on for centos and ubuntu too way too commonly, and I have this feeling we're now making so many changes and running node/ci image builds etc that we roll over all the time => have issues all the time.... just my gut feeling.

@lentzi90 @Sunnatillo WDYT, this needs to be understood and fixed for good :(

Sunnatillo:
> > Just a note that the pre-release pipeline doesn't work with the keyword I used and can be only tested after #1067 is merged.
> > However, I tested the node and CI image building now and it seems to work with this change. They are still running but have passed the phases where the changes from this PR would affect the run:
> > 
> > * replay of node image building with pipeline changes: https://jenkins.nordix.org/view/Metal3/job/metal3-node-image-building/63/
> > * replay of CI image building with pipeline changes: https://jenkins.nordix.org/view/Metal3/job/metal3-ci-image-building/42/
> > 
> > but once again centos is having some issues that are not related to the changes:
> > ```
> > qemu-img: error while reading at byte 6186532864: Input/output error
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This exactly same error occurs on node and CI image building. I have seen same errors before and without doing anything they went away at some point so not sure if there is anything we can do 😕
> 
> Since this is somehow related to the image used, can it simply be that the image where our names point to do not really work, and there is some sort of image length mismatch that goes undetected and it just crashes?
> 
> This is now going on for centos and ubuntu too way too commonly, and I have this feeling we're now making so many changes and running node/ci image builds etc that we roll over all the time => have issues all the time.... just my gut feeling.
> 
> @lentzi90 @Sunnatillo WDYT, this needs to be understood and fixed for good :(

We always point to latest images. Latest image might have some issue and we see that on out CI  immidiately. We can point to exact release and change it time to time. 
This could also be networking issue. But I have seen this issue mostly in Centos. This inclines towards first guest.

jenkins/image_building/build-image.sh
```diff
@@ -16,7 +16,11 @@ trap cleanup EXIT
 # Make sure we run everything in the repo root
 cd "${REPO_ROOT}" || true
 
-export AGENT_OS="${AGENT_OS:-"ubuntu"}"
+if [ -r /etc/os-release ]; then
+  # shellcheck disable=SC1091
+  . /etc/os-release
+  AGENT_OS="${ID}"
+fi
```
lentzi90, line 23: I'm not understanding why we need this. Can we just use `IMAGE_OS`?

lentzi90 (commented):
I'm not sure what to do here. Maybe after OSS EU I will have some time to look deeper into the image builds.

jenkins/image_building/build-image.sh
erjavaskivuori: Because IMAGE_OS might be different than the OS of the agent we are using. On the image_building.pipeline we only use ubuntu agents for building both centos and ubuntu images. So need for the AGENT_OS variable was only introduced after the pre-release pipeline came into the picture because for now we have to use agent with the same OS as the built image has because CAPM3 e2es expect that to be so.

PR #1071, opened 2025-08-25
Pin used CentOS release in image building

erjavaskivuori:
Use spesific CentOS release to improve stability in image building.


jenkins/image_building/build-image.sh
```diff
@@ -66,16 +66,24 @@ if [[ "${IMAGE_OS}" == "ubuntu" ]]; then
...
+  export DIB_LOCAL_IMAGE="${REPO_ROOT}/${centos_image}"
 fi
 
 if [[ "${IMAGE_TYPE}" == "node" ]]; then
   # The default data source for cloud-init element is exclusively Amazon EC2
   export DIB_CLOUD_INIT_DATASOURCES="ConfigDrive"
   export KUBERNETES_VERSION="${KUBERNETES_VERSION:-"v1.33.0"}"
   export CRIO_VERSION="${CRIO_VERSION:-"v1.32.3"}"
-  export CRICTL_VERSION="${CRICTL_VERSION-"v1.33.0"}"
+  export CRICTL_VERSION="${CRICTL_VERSION:-"v1.33.0"}"
```
erjavaskivuori, line 86: This is not related but I thought it would be okay to fix on this PR

erjavaskivuori:
Let's test this one too 
/test metal3-fullstack-build

erjavaskivuori:
> Which one should be merged. You and @Sunnatillo are both working on the same, #1070 vs #1071?

Ohh I didn't know Sunnat was also working on this and didn't check the PRs before creating mine. Sunnat has created the PR first so I think it can be merged. But since there isn't fix for the fullstack building maybe those changes can be left for this PR?

tuminoid:
> > Which one should be merged. You and @Sunnatillo are both working on the same, #1070 vs #1071?
> 
> Ohh I didn't know Sunnat was also working on this and didn't check the PRs before creating mine. Sunnat has created the PR first so I think it can be merged. But since there isn't fix for the fullstack building maybe those changes can be left for this PR?

Sounds like a plan. You should rebase after his PR gets merged and then you can build your fullstack fixes on top.

jenkins/scripts/dynamic_worker_workflow/build_ipa.sh
```diff
@@ -57,6 +57,21 @@ IPA_IMAGE_NAME="${IPA_IMAGE_NAME:-ironic-python-agent}"
 IPA_IMAGE_TAR="${IPA_IMAGE_NAME}.tar"
 IPA_BASE_OS="${IPA_BASE_OS:-centos}"
 IPA_BASE_OS_RELEASE="${IPA_BASE_OS_RELEASE:-9-stream}"
+
+if [[ "${IPA_BASE_OS}" == "centos" ]]; then
+  numeric_release=9
+  date="20250812.1"
```
Sunnatillo, line 63: we could remove "date" variable and directly define the centos_image variable for the sake of reducing varibles. 
Please also check if "numeric_release" variable is needed?
Also, it would be nice if we rename centos_image to centos_upstream_image to make consistent name variables. WDYT?

erjavaskivuori: Thank you for the suggestions, I will make these changes. I think it is a good idea to name it the same, though I noticed it's "centos_upstream_img" in the build-image.sh but anyhow, I will use that to be consistent :smile:

Sunnatillo (commented):
/approve

PR #1074, opened 2025-08-25
Automatically fetch latest pre-release k8s version

erjavaskivuori:
Fetch the latest pre-release Kubernetes version for a specified minor version. This is to require less manual changes.

### DIFF
--- a/jenkins/image_building/build-image.sh
+++ b/jenkins/image_building/build-image.sh
@@ -74,6 +74,16 @@ if [[ "${IMAGE_TYPE}" == "node" ]]; then
   # The default data source for cloud-init element is exclusively Amazon EC2
   export DIB_CLOUD_INIT_DATASOURCES="ConfigDrive"
   export KUBERNETES_VERSION="${KUBERNETES_VERSION:-"v1.33.0"}"
+
+  if [[ "${PRE_RELEASE:-}" == "true" ]]; then
+    # Extract minor version (e.g., "1.34" from "v1.34.0")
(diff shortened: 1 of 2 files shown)


adilGhaffarDev:
> Since the pipeline changes can't be tested by running the tests from here, I have these two runs where I tested the pipeline changes by replacing the pipeline with my changes:

Is this going to be a periodic or just a PR test?
I don't think we need this as periodic. We will run this test only when the k8s release cycle is in the pre-release phase. 

Otherwise nice work
/lgtm

erjavaskivuori:
> Is this going to be a periodic or just a PR test? I don't think we need this as periodic. We will run this test only when the k8s release cycle is in the pre-release phase.

We discussed this elsewhere already, but as a summary; the periodic is already there but we can discuss again if that is needed.

adilGhaffarDev:
> > > Since the pipeline changes can't be tested by running the tests from here, I have these two runs where I tested the pipeline changes by replacing the pipeline with my changes:
> > 
> > 
> > Is this going to be a periodic or just a PR test? I don't think we need this as periodic. We will run this test only when the k8s release cycle is in the pre-release phase.
> > Otherwise nice work /lgtm
> 
> I agree with Adil here, we need this test only when new Kubernetes pre-releases are available. After GA release, we do not have to run. My suggestion to keep it in periodic table, and make it to run manually. Anyhow we need to set minor release version on new test, right?

yes we can keep it in periodic and not run it, I am ok with that.

PR #1076, opened 2025-08-27
Uplift CI images to Ubuntu 24.04

### LINKED ISSUE #954: Build CI image with Ubuntu 24.04 Noble
We have been testing Ubuntu 24.04 Noble manually. It works fine, and it is needed for some features going forward. Change base image of the CI image to Noble, and do any necessary changes to make it work. 

ref: https://github.com/metal3-io/metal3-dev-env/pull/1470

erjavaskivuori:
Uplift CI images to Ubuntu 24.04

Fixes #954

### DIFF
--- a/jenkins/image_building/build-image.sh
+++ b/jenkins/image_building/build-image.sh
@@ -31,7 +31,7 @@ if [[ "${AGENT_OS}" == "ubuntu" ]]; then
   sudo sed -i "s/^#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf > /dev/null || true
 
   sudo apt-get update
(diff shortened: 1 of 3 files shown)


erjavaskivuori:
I tested the build-image.sh script on a vm with ubuntu 24.04 and at least there the script worked. I didn't come across other differences than that qemu package doesn't exist anymore on ubuntu 24.04 but is separated into multiple packages from which it seems that we would need the qemu-system and qemu-utils.

But just a reminder that running the ci-image-building test doesn't assure that the image building script itself will work on 24.04 since on jenkins the CI image we use to build the new image has ubuntu 22.04.

Sunnatillo (commented):
/approve

PR #1084, opened 2025-09-03
Fix clean resources script

Sunnatillo:
Fixes Resource cleanup script which is failing with error "externally-managed-environment".

P.S [Adam]
https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3


Rozzii (dismissed):

tuminoid (commented):
/approve
/retest

PR #1088, opened 2025-09-11
Add prow keywords for k8s upgrade pre-release tests

erjavaskivuori:
Add prow keywords to test k8s upgrade with pre-release versions.


Sunnatillo:
We do not need to run pre release tests on PRs

PR #1097, opened 2025-09-15
Add pods scaling tests to  BML

Sunnatillo:
This PR adds pod scaling tests to BML. Where after pivoting we deploy 650 pods to our target cluster.

### DIFF
--- a/jenkins/scripts/bare_metal_lab/bml_integration_test.sh
+++ b/jenkins/scripts/bare_metal_lab/bml_integration_test.sh
@@ -58,11 +58,21 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook -v "${CI_DIR}"/deploy-lab.yaml
 # In the bare metal lab, we have already cloned metal3-dev-env and we run integration tests
 # so no need to clone other repos.
 if [[ "${REPO_NAME}" == "metal3-dev-env" ]]; then
-    cd "${HOME}/tested_repo"
+    export METAL3_DIR="${HOME}/tested_repo"
 else
-    cd "${HOME}/metal3"
+    export METAL3_DIR="${HOME}/metal3"
 fi
+
+cd "${METAL3_DIR}"
 
 echo "Running the tests"
 
-make test
+make provision
+make pivot
+
+# Run Pods Scaling test
+export ANSIBLE_CONFIG="${CI_DIR}"/tasks/pod_scaling/ansible.cfg
+ANSIBLE_FORCE_COLOR=true ansible-playbook -b "${CI_DIR}"/tasks/pod_scaling/pod-scaling.yaml -i "${CI_DIR}"/tasks/pod_scaling/inventory.ini
+
+make repivot
+make deprovision
--- a/jenkins/scripts/bare_metal_lab/tasks/pod_scaling/pod-scaling.yaml
+++ b/jenkins/scripts/bare_metal_lab/tasks/pod_scaling/pod-scaling.yaml
@@ -1,0 +1,206 @@
+---
+- name: Increase max number of pods on worker node
+  hosts: kube_worker_node
+  gather_facts: false
+  become: true
+
+  tasks:
+  - name: Add maxPod value on worker node kubelet config
+    ansible.builtin.blockinfile:
+      path: /var/lib/kubelet/config.yaml
+      block: |
+        maxPods: 700
+      owner: root
+      group: root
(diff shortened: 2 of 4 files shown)


Sunnatillo:
tests are passed here:
https://jenkins.nordix.org/view/Metal3/job/metal3-bml-integration-test-centos/445/

jenkins/scripts/bare_metal_lab/bml_integration_test.sh
```diff
@@ -58,11 +58,26 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook -v "${CI_DIR}"/deploy-lab.yaml
...
-    cd "${HOME}/metal3"
+    export METAL3_DIR="${HOME}/metal3"
 fi
 
+cd "${METAL3_DIR}"
+
 echo "Running the tests"
 
-make test
+if [[ ${POD_SCALING_TEST} == true ]]; then
```
lentzi90, line 70:
```suggestion
if [[ "${POD_SCALING_TEST}" == "true" ]]; then
```

jenkins/scripts/bare_metal_lab/tasks/pod_scaling/pod-scaling.yaml
```diff
@@ -0,0 +1,206 @@
...
+      kind: Deployment
+      name: nginx-deployment
+      namespace: pod-scaling-test
+      replicas: 650
+      wait_timeout: 1200
+      kubeconfig: /home/metal3/.kube/config
+
+  - name: Pause for 5 minutes to run 650 pods simultaneously
+    ansible.builtin.pause:
+      minutes: 5
```
lentzi90, line 188: I think we need to do more than just pausing here. We should check that the Kubernetes API is stable, that the nodes are healthy and that the pods are running.

lentzi90 (commented):
This achieves the goal and in that sense it is ok. I am fine taking it in as an initial implementation, but I think we are now at a point where we need to think about how these BML tests should work.

We are treating the nodes as pets (there is always exactly 1 worker and 1 control-plane with fixed IPs known beforehand). It doesn't feel quite right to me that we use ansible to go in and edit the kubelet config and restart the kubelet on the node. We could do this through CAPI manifests instead. We could also get the kubeconfig and use it remotely instead of going into the control-plane node itself to access the API. That way we would also ensure that the control-plane endpoint is working, not just the local API endpoint on the node.

Maybe it is time to design a new BML test suite without going through metal3-dev-env? We just need the bootstrap part anyway.

jenkins/scripts/bare_metal_lab/tasks/pod_scaling/pod-scaling.yaml
Sunnatillo: I get your point. But this is going to mess up metal3 dev env repo again. My goal is to not touch other repos. We already have to many things put in metal3 dev env.

Sunnatillo:
> This achieves the goal and in that sense it is ok. I am fine taking it in as an initial implementation, but I think we are now at a point where we need to think about how these BML tests should work.
> 
> We are treating the nodes as pets (there is always exactly 1 worker and 1 control-plane with fixed IPs known beforehand). It doesn't feel quite right to me that we use ansible to go in and edit the kubelet config and restart the kubelet on the node. We could do this through CAPI manifests instead. We could also get the kubeconfig and use it remotely instead of going into the control-plane node itself to access the API. That way we would also ensure that the control-plane endpoint is working, not just the local API endpoint on the node.
> 
> Maybe it is time to design a new BML test suite without going through metal3-dev-env? We just need the bootstrap part anyway.

I get your point. But this is going to mess up metal3 dev env repo again. My goal is to not touch other repos. We already have to many things put in metal3 dev env.

lentzi90:
> I get your point. But this is going to mess up metal3 dev env repo again. My goal is to not touch other repos. We already have to many things put in metal3 dev env.

Sure, I was more thinking of implementing something here in project-infra and just abandon metal3-dev-env. But I think we can start with what you have in this PR then we can refactor later.

lentzi90 (commented):
/approve
I am ok taking this in as a starting point, but as mentioned above, I think we need to improve on it. We should not be using ansible to go in and manage processes on the nodes or treat the nodes as fixed, well-known entities.
We need to prioritize cleaning this up so we don't end up with hard-to-maintain jobs.
/cc @kashifest

Issue #1081, opened 2025-09-01
Build and test Centos 10 node images


kashifest:
Lets build and test Centos 10 node images

PR #1101, opened 2025-09-17
Build CentOS 10-stream node image

kashifest:
Fixes: #1081


kashifest:
ok I digged into this by going inside the test vm. 

I see the following error in cloud-init log

```
Error: Connection activation failed: No suitable device found for this connection (device enp1s0 not available because profile is not compatible with device (mismatching interface name)).
Created symlink '/etc/systemd/system/multi-user.target.wants/keepalived.service' → '/usr/lib/systemd/system/keepalived.service'.
Job for keepalived.service failed because the control process exited with error code.

```

Also because of this interface name mismatch issue keepalived errored out and hence kubeapi server never came up. Initially we think this is a change introduced in CentOS stream 10 and perhaps it would require this PR to merge first https://github.com/metal3-io/metal3-dev-env/pull/1561/ to fix the current issue.

Sunnatillo:
> are we going to stop building CS9 node images?

yes. After this gets merged we switch building centos 10 images.

tuminoid (commented):
/lgtm
/unhold

PR #1108, opened 2025-09-25
Add prow job to generate CR documentation in BMO; and revert 'Install…

MahnoorAsghar:
… podman in basic-checks container'


prow/config/jobs/metal3-io/baremetal-operator.yaml
```diff
@@ -86,6 +86,20 @@ presubmits:
         - make
         image: quay.io/metal3-io/basic-checks:golang-1.24
         imagePullPolicy: Always
+  - name: generate-api-docs
+    branches:
+    - main
+    run_if_changed: '(config/base/crds/bases/.*)$*'
```
tuminoid, line 92: Also needs to run if the makefile changes (could change the make target)

prow/config/jobs/metal3-io/baremetal-operator.yaml
```diff
@@ -86,6 +86,20 @@ presubmits:
...
+    decorate: true
+    spec:
+      containers:
+      - args:
+        - --resources=config/base/crds/bases/
+        - --output=docs/api.md
+        command:
+        - crdoc
+        image: ghcr.io/fybrik/crdoc:latest
```
tuminoid, line 101: Must be pinned to a sha with tag in comment. For example:

```suggestion
        image: ghcr.io/fybrik/crdoc:shashashasha # v0.6.2
```

tuminoid (commented):
This will generate the doc, but its missing the part where it detects if something has changed. That is easier to implement if you look at the existing hack/ scripts how they deal with local run and the prow run.

MahnoorAsghar:
> This will generate the doc, but its missing the part where it detects if something has changed. That is easier to implement if you look at the existing hack/ scripts how they deal with local run and the prow run.

@tuminoid I see that the pattern in hack/ is to launch a podman container, and once in it we execute a command, but I didn't understand what you meant by the detecting if something has changed part? Does the 'run_if_changed' prow argument not take care of that?
(I can ofcourse add a script to the hack/ directory for consistency or benefits)

tuminoid:
> > This will generate the doc, but its missing the part where it detects if something has changed. That is easier to implement if you look at the existing hack/ scripts how they deal with local run and the prow run.
> 
> @tuminoid I see that the pattern in hack/ is to launch a podman container, and once in it we execute a command, but I didn't understand what you meant by the detecting if something has changed part? Does the 'run_if_changed' prow argument not take care of that? (I can ofcourse add a script to the hack/ directory for consistency or benefits)

The point of crdoc generation in tests is to see if rerunning it produces a diff, which would mean the author did not include the updated documentation in their PR.  So it needs to fulfill two things:
1. allow developer to generate the updated API docs and add it to their PR
2. re-generate the API docs in prow, and FAIL if there was any change

tuminoid:
/hold
until https://github.com/metal3-io/baremetal-operator/pull/2605 is completed

PR #1116, opened 2025-10-03
Fix OpenSUSE image OS

nuhakala:
Fixes the `IMAGE_OS` value.

### DIFF
--- a/jenkins/jobs/image_building.pipeline
+++ b/jenkins/jobs/image_building.pipeline
@@ -39,7 +39,7 @@ pipeline {
         axes {
           axis {
             name 'IMAGE_OS'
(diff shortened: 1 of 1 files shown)


Sunnatillo (approved):
/approve

nuhakala:
> We still have `opensuse-leap` in AGENT_OS, is that intended?

Yes. I checked the variables and AGENT_OS comes from `/etc/os-release` file and IMAGE_OS is just a value assigned by me. At this point it would have been maybe more clear to use opensuse-leap value everywhere, but then I would need to to change CAPM3 and dev env as well.

tuminoid (commented):
/lgtm

PR #1117, opened 2025-10-08
:seedling: Allow downloading artifacts to allow pre-preparing directories for image-building

peppi-lotta:
**What this PR does / why we need it**: A docker image building requires that all the element needed for building are in the same directory or sub directory where the Dockerfile is located. This is why I created a [PR ](https://github.com/metal3-io/cluster-api-provider-metal3/pull/2826)that pres FKAS for image building by moving the required CAPM3 directory into the FKAS direcory before building. Because we are using a reusable workflow for image building the directory prep has to be done in a separate job. It seems changes are not retained accross jobs even when they are run in the same workflow so I'm suggesting to use the upload-archive and download-archive actions to pass the pre prepaped FKAS folder from one job to another. Related PR in project-infra can be found here:


.github/workflows/container-image-build.yml
```diff
@@ -102,13 +107,21 @@ jobs:
         echo "IMAGE_TAGS=${IMAGE_TAGS}" >> "${GITHUB_ENV}"
         echo "IMAGE_VERSION=${IMAGE_VERSION}" >> "${GITHUB_ENV}"
 
+    - name: Download artifact
+      id: download-artifact
+      if: ${{ inputs.artifact-name }} != ""
+      uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0
```
tuminoid, line 113: we need a human readable comment about the tag version

Issue #1100, opened 2025-09-17
Secure node images


lentzi90:
We are currently doing a few things with the node images that aren't ideal. We should fix these things so that the images are more generally useful and acceptable.

1. We permit root login. This is not needed, just remove it. See https://github.com/metal3-io/project-infra/blob/d43b5acf6591399c7975c1e72271ca0a30915850/jenkins/image_building/dib_elements/centos-node/post-install.d/90-pre-pull-images#L9 and https://github.com/metal3-io/project-infra/blob/d43b5acf6591399c7975c1e72271ca0a30915850/jenkins/image_building/dib_elements/ubuntu-node/install.d/55-install#L5
2. We use the "devuser" element to create the `metal3ci` user and bake an authorized ssh key for it. This is trickier. It can be useful if cloud-init fails, to have this way of accessing the node, baked in the image. In general we should not need it though. We can still get serial logs that should tell us what is wrong when cloud-init fails, and if all else fails we can always build a custom image with hard-coded credentials. I suggest we remove this from the node images. See https://github.com/metal3-io/project-infra/blob/d43b5acf6591399c7975c1e72271ca0a30915850/jenkins/image_building/build-image.sh#L61-L63

PR #1118, opened 2025-10-08
Secure node images by removing root login and devuser element

lentzi90:
This commit does the following:

1. Remove PermitRootLogin configuration from CentOS and Ubuntu node images to prevent root login access. This is not used nor needed.
2. Remove the devuser element dependency from all node image types (centos-node, ubuntu-node, leap-node) to eliminate the hard-coded metal3ci user with baked-in SSH keys. While sometimes useful, this makes the images unusable for other things where metal3ci should not have access.
3. Update documentation (README files) to reflect the removal of devuser dependencies

Fixes: #1100


lentzi90:
Centos passed image verification. Ubuntu failed building the image with some strange error that I do not understand. Leap is broken on main so I am disregarding that one for now.
Rerunning to see if the ubuntu error is persistent.
/test metal3-node-image-building

Rozzii:
/test metal3-node-image-building
Ubuntu package management was flaky in other tests too let's try again

PR #1125, opened 2025-10-15
Fix missing devuser variables

lentzi90:
This was mistakenly removed when removing the devuser element from node image elements. The variables are still needed for the CI images.


smoshiur1237:
I think only Centos based package name would be enough. But let see the test ...

nuhakala:
> @nuhakala do you have any idea?

I checked logs. The image building pipeline is not using the fix I made here https://github.com/metal3-io/project-infra/pull/1119 and hence the error. No idea why the fix is not used.

lentzi90:
Pushed a second commit to do the same fix as in https://github.com/metal3-io/project-infra/pull/1119 for Leap CI images.
/test metal3-ci-image-building

PR #1126, opened 2025-10-16
Uplift go 1.24.8 for PROW

smoshiur1237:
Uplift go 1.24.8


smoshiur1237:
/cc @kashifest @lentzi90 
Please also override the mandatory test

PR #1130, opened 2025-10-27
🌱 Update the osv-scanner workflow to accept the branch name

adilGhaffarDev:
Update the osv-scanner workflow to accept the branch name. 
The default branch is main, and this can be set from the repo that will be using this workflow.


tuminoid (commented):
/approve

PR #1142, opened 2025-11-04
add release-1-11 jobs to project infra tests

tuminoid:
We have forgotten to also add release-1-11 tests into project-infra's own test set.


kashifest:
/approve
@Rozzii if this was missing in the prior art, we need to add.

PR #1146, opened 2025-11-05
Change jenkins pipeline files extentions to .groovy

Sunnatillo:
This PR change jenkins pipeline files extentions to .groovy which makes easier to read and use groovy linters in IDE.


lentzi90:
Alternatively, if that is not possible because it fails validation, let's then fix the formatting first without renaming.

Sunnatillo:
@lentzi90 @tuminoid I have changed only filenames on this PR. We should merge cicd PR and this one at the same time.
https://gerrit.nordix.org/c/infra/cicd/+/24553

PR #1148, opened 2025-11-05
fix few Prow config issues/typos

tuminoid:
Fixing couple Prow config issues, like typos, missing branches, removing locked branches etc.


Sunnatillo (commented):
/lgtm

PR #1149, opened 2025-11-06
:seedling: Add test for quick start guide

peppi-lotta:


hack/quick-start-test/bmo/ironic.env
```diff
@@ -0,0 +1,3 @@
+DEPLOY_KERNEL_URL=http://192.168.222.1:6180/images/ironic-python-agent.kernel
+DEPLOY_RAMDISK_URL=http://192.168.222.1:6180/images/ironic-python-agent.initramfs
+IRONIC_ENDPOINT=https://192.168.222.2:6385/v1/
```
nuhakala: `https` but you have not configured tls for ironic

nuhakala (commented):
When I tried to deploy this, bmo was not able to connect to ironic and inspection did not even start. there was error `microversions not supported by endpoint` in bmo log, which I have really no clue what it means. Something off with network.

I tried to enable tls if that would fix it, but it did not. I still got the same error. Also, I suspect that the bmo kustomization is not configured properly for tls, the bmo deployment did not have reference to ironic certs. I think [this](https://github.com/metal3-io/baremetal-operator/blob/main/config/components/tls/tls_ca_patch.yaml) patch should be applied if tls were to be enabled.

PR #1156, opened 2025-11-20
Add k8s indentation rules for yamllint

smoshiur1237:
The rule to follow k8s indentation are added and also added optional line-length check. It will catch the following issue in new pull request
```
# Allowed
foo:
- bar

# not Allowed
foo:
  - bar
```


tuminoid (commented):
/approve

PR #1160, opened 2025-11-24
add main branch to be scanned in metal3 osv-scanner pipeline

tuminoid:
Add main branch to be scanned in metal3 osv-scanner pipeline.

We'll remove osv-scanner Github workflows, so main has to be scanned by the Jenkins job as well.


Rozzii (commented):
/lgtm

PR #1165, opened 2025-11-27
🌱 Minor improvements in osv-scanner pipline

adilGhaffarDev:
Improvements:

- Now in addition collecting the osv-scnner results in artifacts they are printed in jenkins logs too.
- osv-scanner step is red now when there is failure.

Here is how it looks like:
https://jenkins.nordix.org/blue/organizations/jenkins/metal3-osv-scanner-metal3-repos/detail/metal3-osv-scanner-metal3-repos/20/pipeline


tuminoid (commented):
/approve

Thanks  @adilGhaffarDev !

PR #1166, opened 2025-11-27
bump golang 1.24.10 for basic-checks

tuminoid:
This bump is not per se required, but it works as a test commit for image SBOM work as that relies on Github OIDC tokens that are only granted for actual workflow runs and hence cannot test the SBOMs offline.

This will be "test" for container signing PR (TBD).


Rozzii:
#1164 has been merged
/lgtm

PR #1169, opened 2025-11-27
fix build image permissions for signing

tuminoid:
We need id-token in the container-image-build workflow, but it must inherit it from the calling workflow.


Sunnatillo (commented):
/approve

PR #1178, opened 2025-12-01
🌱 amend Lychee config

tuminoid:
Edit Lychee config:
- no-progress = don't print progress
- exclude-all-private = private addresses are unreachable anyways
- insecure = ignore ssl problems, not our problem

Left for future improvements (finds a lot of broken anchors...)
- include-fragments = check anchor links within document

PR #1188, opened 2025-12-04
amend Lychee config for PR checks as well

tuminoid:
I forgot Lychee has two reusable jobs, one for schedued runs (fixed in #1178) and one for PR runs (this fix).


lentzi90 (commented):
/approve

PR #1190, opened 2025-12-05
make dev-env tests in ironic-image use PR content

tuminoid:
Currently, running dev-env-integration tests in ironic-image do not use PR contents, making it useless test. Export the right config to make it build and use PR content.


adilGhaffarDev (approved):
/lgtm

PR #1194, opened 2025-12-08
reduce Lychee concurrency to avoid rate limiting

tuminoid:
Fix Lychee config in multiple ways to improve its reliability:
- make user-agent a Brave clone (privacy protecting browser gets us maybe some slack)
- properly export GITHUB_TOKEN (I was wondering why we get rate limited on Github pages too, with token it should be "unlimited")
- reduce max concurrency to 8 (from 128, another source of surely getting rate limited)
- allow Lychee to retry after timeout, if something wonky happens

Ref: https://github.com/lycheeverse/lychee#commandline-usage

This makes it possible for large PRs with many URLs to be checked like https://github.com/metal3-io/metal3-docs/pull/636 to pass.


adilGhaffarDev (approved):
/lgtm

PR #1198, opened 2025-12-11
add re-approve workflow for github workflows

tuminoid:
Add the re-approver workflow so the workflows re-run automatically if ok-to-test is present. This affects non-member contributors.


lentzi90 (commented):
/approve

PR #1199, opened 2025-12-11
🐛 Fix osv-scanner to also check pre-releases

adilGhaffarDev:
- Fix osv-scanner to also check pre-releases
 - Check last 2 branches of irso instead of 3
tested here: https://jenkins.nordix.org/blue/organizations/jenkins/metal3-periodic-osv-scanner-metal3-repos/detail/metal3-periodic-osv-scanner-metal3-repos/17/pipeline/


Rozzii (commented):
/lgtm

PR #1207, opened 2025-12-18
Bump k8s version to v1.35.0

Sunnatillo:
Uplifts k8s and CRICTL_VERSION version to v1.35.0


kashifest (approved):
Now all the tests are green

PR #1213, opened 2026-01-05
Bump CRIO to v1.35.0

kashifest:
Bumps CRIO to v1.35.0


adilGhaffarDev (commented):
/lgtm
we need to do it in cicd too.

PR #1218, opened 2026-01-12
Fix pre-release test to use ubuntu agents only

kashifest:
Hardcoding the agents to use ubuntu only as it is for image building also. This should not affect the image we are building and testing. This is only affecting the build env. CentOS agents are missing packages and failing to build the image currently.

### DIFF
--- a/jenkins/jobs/pre_release_tests.groovy
+++ b/jenkins/jobs/pre_release_tests.groovy
@@ -20,9 +20,9 @@ script {
     echo "Checkout ${ci_git_url} branch ${ci_git_branch}"
 
     if ( "${GINKGO_FOCUS}" == 'integration' ) {
-        agent_label = "metal3ci-8c16gb-${IMAGE_OS}"
+        agent_label = 'metal3ci-8c16gb-ubuntu'
     } else if ( "${GINKGO_FOCUS}" == 'k8s-upgrade' ) {
-        agent_label = "metal3ci-8c24gb-${IMAGE_OS}"
+        agent_label = 'metal3ci-8c24gb-ubuntu'
     }
 }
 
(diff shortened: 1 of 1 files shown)


smoshiur1237 (commented):
/lgtm

kashifest:
> for centos test its trying to find Ubuntu image. The reason is in our tests we set image_os according to test environment(ci image). https://github.com/metal3-io/cluster-api-provider-metal3/blob/33576e36f6c60ca47d5d08458b3c4bdf82c33d8c/scripts/environment.sh#L20 I think thats why Erja did it like that.

There’s something missing here. It has successfully built CentOS image but the IMAGE_OS is not being passed correctly to e2e tests. I need to check how we do it for node image building

lentzi90 (commented):
I think it makes sense like this. When CAPM3 e2e can work without metal3-dev-env, we can switch this also to run the CAPM3 e2e basic test or something like that. For now I do not think it is worth the effort.
/approve

adilGhaffarDev:
> So now both the test passed. If this lands as is, the prerelease integration test as well as the periodic will pass. However theres another issue. It wont pass the upgrade test since I am changing to run dev-env test from e2e now. Let me clarify.
> 
> In this PR I changed script which verifies the image built to `jenkins/image_building/verify-node-image.sh`. It is the same script we use also for node image building which is a dev-env test.
> 
> The existing test here was the e2e test `jenkins/scripts/dynamic_worker_workflow/e2e_tests.sh`. This was not working here since the image built here specially the centOS was failing due to the fact that the Image variables are not passed in the most entangled setup project-infra -> dev-env -> CAPM3. For this to work, we need to change CAPM3 e2e to accomodate project-infra dictating those variables.
> 
> I am not sure if it is worth the effort to now go fix CAPM3 for this to work or no. I know upgrade test for pre-release also has some value but I am up for opinions since it would take effort to fix this as is. Other easier way would be we only keep Ubuntu based test and keep integration and upgrade using e2e as is and drop CentOS test.


I am in favor of removing centos for now and using capm3 e2e tests. If we really want to test centos also in the future, lets create an tracking issue. This way we will have both integration and k8s upgrade tests. 
If we switch to `jenkins/image_building/verify-node-image.sh` we can only test integration.

lentzi90:
> I am in favor of removing centos for now and using capm3 e2e tests. If we really want to test centos also in the future, lets create an tracking issue. This way we will have both integration and k8s upgrade tests.
If we switch to jenkins/image_building/verify-node-image.sh we can only test integration.

I am fine with this also. Is it so that we then upgrade from a stable release to the new pre-release? That would be nice!

Sunnatillo (commented):
/lgtm

Supporting upgrade tests can be handled separately.

adilGhaffarDev:
Oh I checked we dont have k8s upgrade test right now for pre-release, I thought we already have it and this PR is going to break it. 
I agree k8s can done in the future if needed. 

Lets do one thing in followup PR, lets change the name to metal3-ubuntu-dev-env-integration-k8s-pre-release-test-main

PR #1219, opened 2026-01-12
🌱 Fix k8s version for in-place upgrade tests

adilGhaffarDev:
Fix k8s version for in-place upgrade tests
make sure to get k8s version from jjb
follow up of this: https://github.com/metal3-io/project-infra/pull/1215


Sunnatillo (commented):
/approve

PR #1220, opened 2026-01-13
Move groovy lint test from github actions to prow

Sunnatillo:
This PR moves running groovy lint test from github actions to prow. 
Also changes allow to run same test locally.


tuminoid (commented):
Nits.

Also Q: since we add this script, should we use it instead of GH action to lint to groovy as well? now we need to maintain two. This looks pretty much Prow ready, so consider removing the pr-groovy-lint workflow and add Prow test instead.

tuminoid (commented):
/approve

PR #1224, opened 2026-01-21
Prow: Bump container images

lentzi90:
This is to get a fix for a bug that is causing the dashboard to not render properly.

Changes already applied.


tuminoid (commented):
/approve

PR #1239, opened 2026-02-05
increase blunderbuss reviewer count to 3

tuminoid:
We often end up in situation that both of the reviewers randomized by Blunderbuss are more or less inactive, leading to almost no-one actually reviewing some PRs. Increase the number from 2 to 3.


lentzi90 (commented):
/approve

PR #1243, opened 2026-02-10
:bug: Improve venv setup for ostack resource cleanup

Rozzii:
This PR:
 - makes clean_resource.sh utilize pre-installed openstack client
 - modifies ubuntu ci image building to include openstack cli venv
   bootstrapping
 - bump the DIB version from 3.33.0 (2+years old) to most recent 3.40.2

This commit is needed:
Because there have been situations where the DEB repository list was
out of date on a in-use ubuntu ci image thus the python3.12-venv package
install failed on the ci image while running clean_resources.sh.

Also it doesn't make sense to bootstrap the venv for Openstack client during
the cleanup job, when the venv can be bootstrapped during ci image building.

DIB bump was needed because 3.33.0 was struggling with some of its
dependencies, at least with 'pkg_resources' (that made it fail so not sure if
there was any other problematic package other than this).

### DIFF
--- a/jenkins/scripts/clean_resources.sh
+++ b/jenkins/scripts/clean_resources.sh
@@ -43,18 +43,21 @@ cleanup() {
     done
 }
 
-sudo apt install -y python3.12-venv
+WORK_VENV="${HOME}/civenv"
+WORK_VENV_ACTIVATOR="${WORK_VENV}/bin/activate"
 
-rm -rf venv
-python3 -m venv venv
+if [[ ! -x "${WORK_VENV_ACTIVATOR}" ]]; then
+    sudo apt-get update
+    sudo apt-get install -y python3.12-venv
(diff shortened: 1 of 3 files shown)


smoshiur1237 (commented):
/lgtm

smoshiur1237 (commented):
Now it is moving the image building job.
/lgtm

jenkins/scripts/clean_resources.sh
```diff
@@ -72,3 +75,5 @@ cleanup
 
 #unset openstack variables
 unset "${!OS_@}"
+# deactiveate the python venv (for non ci use)
```
tuminoid: This doesn't matter, venv variables exist only in the context sourcing them, ie. they disappear when the script exits.

smoshiur1237 (commented):
/lgtm

PR #1247, opened 2026-02-12
Prow: S3 lifecycle policy for log retention

lentzi90:
Add S3 Lifecycle policy for removing logs that are older than 90 days. We only need this for the logs bucket since tide and status-reconciler doesn't accumulate data.

Also update DNS documentation with information about self-management through Netlify.

### DIFF
--- a/prow/README.md
+++ b/prow/README.md
@@ -55,12 +55,14 @@ A webhook must be configured in GitHub to
 send events to Prow and a DNS record must be configured so that
 <https://prow.apps.test.metal3.io> points to the IP where Prow can be accessed.
 
-The DNS records are managed by CNCF for us. Any changes to them can be done
-through the
-[service desk](https://cncfservicedesk.atlassian.net/servicedesk/customer/user/login?destination=portals).
-All project maintainers have (or can request) access. See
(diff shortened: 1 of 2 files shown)


smoshiur1237 (commented):
/lgtm

prow/README.md
```diff
@@ -55,9 +55,11 @@ A webhook must be configured in GitHub to
...
 <https://prow.apps.test.metal3.io> points to the IP where Prow can be accessed.
 
-The DNS records are managed by CNCF for us. Any changes to them can be done
-through the
-[service desk](https://cncfservicedesk.atlassian.net/servicedesk/customer/user/login?destination=portals).
+The DNS records are managed through
+[Netlify](https://app.netlify.com/teams/cncf-projects/dns/metal3.io). CNCF
+controls the Netlify project. Maintainers can request access through the
+[service
+desk](https://cncfservicedesk.atlassian.net/servicedesk/customer/user/login?destination=portals).
```
kashifest: We as maintainers can self manage the DNS records, I think your statements are correct but I was thinking if we should also clarify this.

PR #1250, opened 2026-02-13
Revert "Comment out CAPM3 build job"

lentzi90:
Reverts metal3-io/project-infra#1249

Let's put the job back and see if we still get constantly re-triggered jobs or if they now behave as they should...
/hold


Rozzii (commented):
/approve

PR #1251, opened 2026-02-16
attempt to fix prow test trigger for capm3 build

Rozzii:
This PR:
 - Changes the regex trigger to a more accurate regex

This commit is needed because prow is periodically re-triggering this test even when it is not needed.


smoshiur1237 (commented):
/lgtm

PR #1252, opened 2026-02-17
improve Jenkins resource cleanup script

Rozzii:
This PR:
 - Makes sure Jenkins only cleans up OpenStack VMs that were created by the Nordix Jenkins instance.

This commit was needed because otherwise the cleanup script only takes into consideration the creation time and the name prefix of the VM thus even persistent VMs that have the prefix would be cleaned up. The goal of this script is to remove resources dynamically created by Jenkins and nothing else.


Rozzii:
irrelevant

kashifest (commented):
/approve

PR #1258, opened 2026-02-23
switch jenkins jobs to use jnlp agent

Rozzii:
This PR:
 - Modifies the Jenkins pipeline scripts to use the agent templates with jnlp.
 - Changes templates that were previously 8c16gb to 4c16gb.

Note: previously used non jnlp templates with 8c16gb names were still using 4c16 OpenStack flavors, their name is incorrect.

Note 2: old agent templates were using ssh connection that required allocating IPs and ports that were publicly available on the internet, jnlp based agents will utilize a callback mechanism that does not require public IPs and ports.

### DIFF
--- a/jenkins/jobs/bmo_e2e_optional_tests.groovy
+++ b/jenkins/jobs/bmo_e2e_optional_tests.groovy
@@ -14,7 +14,7 @@ pipeline {
     stages {
         stage('Run Baremetal Operator optional e2e tests') {
             matrix {
-                agent { label 'metal3ci-8c16gb-ubuntu' }
+                agent { label 'metal3ci-4c16gb-ubuntu-jnlp' }
                 axes {
                     axis {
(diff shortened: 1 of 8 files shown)


Rozzii:
So these get all the agents properly allocated but seem to time out almost at the end of the tests:
 - https://jenkins.nordix.org/view/Metal3/job/metal3-fullstack-build/65/  - This time out while the test is still running almost at the end
 - https://jenkins.nordix.org/view/Metal3/job/metal3-ci-image-building/90/ - this times out but only on the leap verification's end, the other images finish the verification
 
 It is not clear to me if it is just general slowness or related to the fact that these are new agents so Jenkins needed more time to instantiate them but other than the test taking a bit longer there are no fundamental failures I would say let's merge this and if we need we can tune timeouts or investigate what could cause slow VM provisioning if that is the case.

PR #1267, opened 2026-02-27
Fix manifests for BML test

Sunnatillo:
Add required api and kind for ipAddressFromIPPool and prefixesFromIPPool  for BML test manifests


adilGhaffarDev (commented):
/lgtm

PR #1275, opened 2026-03-04
Revert "jenkins: set PATH in the stage's environment block"

fmuyassarov:
This reverts commit e23ead1ce9f5a578e4cba87ee0071dc009fa347f as we have a "proper" fix that was made outside of git on the Jenkins to include the -i flag to ` sudo -u foo /usr/bin/java`.


Sunnatillo (commented):
/approve

PR #1276, opened 2026-03-05
Add rsa ssh key for Oracle cloud machines

Sunnatillo:
Oracle jenkins plugin does not support RSA keys yet


smoshiur1237 (commented):
/lgtm

PR #1281, opened 2026-03-06
Rename KUBERNETES_VERSION_FROM to KUBERNETES_VERSION_UPGRADE_FROM

Sunnatillo:
CAPI uses KUBERNETES_VERSION_UPGRADE_FROM variable in scaling tests. By renaming KUBERNETES_VERSION_FROM to KUBERNETES_VERSION_UPGRADE_FROM we reduce one more var name.

Depends on:
cicd PR: https://gerrit.nordix.org/c/infra/cicd/+/25228

After current PR we can test: 
CAPM3 PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/3211


Sunnatillo:
> > > Is there a test we can run here?
> > 
> > 
> > No, because we are changing in three different places. We need to take them in around the same time.
> 
> Yeah, I was afraid of that. We already merged some in CAPM3 main and branches broke, we fixed it here and there and now it goes off again the same :)

Luckily it is not big change, less probability of failure. We can merge cicd and project infra, then test CAPM3 PR.

kashifest (commented):
/approve

PR #1286, opened 2026-03-09
Set correct k8s upgrade tests in prow

Sunnatillo:
Updates k8s upgrade tests to latest configuration in prow


lentzi90 (commented):
/approve

PR #1302, opened 2026-03-26
Make leap ci image to run in OCI cloud

Sunnatillo:
This PR:
1. Makes networkManager default
2. Intalls virtio drivers to support paravirtualized mode of OCI
Paravirtualized mode requires virtio devices to properly boot image in OCI.
3. Set hostonly=no so the initrd is generic and  ensures the image boots
correctly on any target
4. Adds console parameters to ensure the image boots properly in OCI


tuminoid (commented):
/approve

Rozzii (commented):
/lgtm

PR #1317, opened 2026-04-13
ignore otel 1.43 bump for release branches in osv-scanner job

tuminoid:
[otel 1.43 bump has vuln fix for BSD only](https://osv.dev/vulnerability/GHSA-hfvc-g4fc-pqhx), but it requires go 1.25 so we cannot bump it for no value in release branches. Just silence the vuln for branches. It has been bumped in main.

Silencing will make the osv-scanner job valuable again, otherwise it'll fail until 1.12 falls out of maintenance, which is 5 months away.

Global config file applies to all submodule go.mod files. If the config file would be placed in the repo and auto-discovered, each ignore would apply only to go.mod next to it. In CAPM3 case, that'd be 3 files in both release branches. Putting it in the global ignore solves the whole thing with one PR.

### DIFF
--- a/jenkins/jobs/osv_scanner_metal3.groovy
+++ b/jenkins/jobs/osv_scanner_metal3.groovy
@@ -62,6 +62,22 @@ def runOsvScan = { String repoName, String refType, String ref, String repoUrl, 
         try { gv = sh(script: 'make go-version', returnStdout: true).trim() } catch (e) { echo "make go-version failed: ${e}" }
         gv = gv ?: goVersion
         sh "echo 'GoVersionOverride = \"${gv}\"' > config.toml"
+        // GHSA-hfvc-g4fc-pqhx: BSD/Solaris-only PATH hijack in otel kenv; fix requires otel v1.43.0 (Go 1.25).
+        // Only affects release branches where otel cannot be bumped due to Go version constraint.
+        def ignoredBranches = [
+            'CAPM3': ~/^release-1\.(1[0-2]|[0-9])$/,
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

kashifest:
> otel 1.43 bump is for BSD only, but it requires go 1.25 so we cannot bump it for no value in release branches. Just silence the vuln for branches. It has been bumped in main.
> 
Its for BSD only? https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.43.0 this changelog seems to have nonbsd change as well? or am I looking at the wrong change log?

tuminoid:
> > otel 1.43 bump is for BSD only, but it requires go 1.25 so we cannot bump it for no value in release branches. Just silence the vuln for branches. It has been bumped in main.
> 
> Its for BSD only? https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.43.0 this changelog seems to have nonbsd change as well? or am I looking at the wrong change log?

The vuln is in kenv usage which is bsd-only. v1.43.0 for sure has plenty of other changes, including bumping to go 1.25 etc.

https://osv.dev/vulnerability/GHSA-hfvc-g4fc-pqhx

> The fix for GHSA-9h8m-3fm2-qjrq (CVE-2026-24051) changed the Darwin ioreg command to use an absolute path but left the BSD kenv command using a bare name, allowing the same PATH hijacking attack on BSD and Solaris platforms.

tuminoid:
Updated the description with better wording and the link, it was vague.

kashifest (commented):
/lgtm

PR #1320, opened 2026-04-21
Fix agent label for pre-release test

kashifest:
After moving the test to OCI we only have one type of agent which is 8C32gb- . Fixing it for pre-release test since it is now stuck finding the wrong agents.


Rozzii (commented):
/approve

tuminoid (commented):
/lgtm

PR #1321, opened 2026-04-22
add docker server vulns into osv-scanner's ignore

tuminoid:
They are in server side auth plugins, we use client in tests. No fix available, as it'd needs switch to moby/moby module.


kashifest (commented):
/approve

Sunnatillo (commented):
/lgtm

PR #1323, opened 2026-04-23
Add cleanup step in fullstack pipeline

Sunnatillo:
We might reuse jenkins agents some edge cases. Test folders/artifacts needs to be cleaned after it finishes/fails

### DIFF
--- a/jenkins/jobs/dynamic_fullstack_building.groovy
+++ b/jenkins/jobs/dynamic_fullstack_building.groovy
@@ -92,5 +92,12 @@ pipeline {
                 archiveArtifacts "logs-${env.BUILD_TAG}.tgz"
             }
         }
+        cleanup {
+            script {
+                timestamps {
+                    sh './jenkins/scripts/dynamic_worker_workflow/run_clean.sh'
+                }
+            }
+        }
     }
 }


lentzi90 (commented):
/lgtm

tuminoid:
> We might reuse jenkins agents some edge cases. Test folders/artifacts needs to be cleaned after it finishes/fails

Which edge cases? I'd really love to the other way and remove any cleaning related to these VMs.

Sunnatillo:
> > We might reuse jenkins agents some edge cases. Test folders/artifacts needs to be cleaned after it finishes/fails
> 
> Which edge cases? I'd really love to the other way and remove any cleaning related to these VMs.

Oracle jenkins cloud plugin does not have RETENTION_NUMBER variable which allows to set how many times agent will be used. For openstack plugin we set it to 1. 
Closes workaround was to set timeout of the idle agents to 1 minutes. During these window if some other test picks up, agent will be reused.

Rozzii (commented):
I am out of the loop a bit, but is this cleanup script compatible with OCI? Last time I used this script it was written specifically for cleaning up OpenStack VMs.

Sunnatillo:
> I am out of the loop a bit, but is this cleanup script compatible with OCI? Last time I used this script it was written specifically for cleaning up OpenStack VMs.

This  is about cleaning up the VM after running the tests on it.

tuminoid (commented):
/approve

Sigh. We're rolling back all the nice things we did for OCI.

PR #1337, opened 2026-05-21
Fix cleaning for fullstack build

terror96:
Default to Ubuntu in case `OS_IMAGE` is not defined when running clean up. This fixes the `IMAGE_OS: unbound variable` error.

Specify `DEV_ENV_REPO_LOCATION` as the destination directory when cloning metal3-dev-env, so the clone lands in the expected location.

### DIFF
--- a/jenkins/scripts/dynamic_worker_workflow/run_clean.sh
+++ b/jenkins/scripts/dynamic_worker_workflow/run_clean.sh
@@ -2,6 +2,8 @@
 
 set -eux
 
+IMAGE_OS="${IMAGE_OS:-ubuntu}"
+
 if [[ "${IMAGE_OS}" == "ubuntu" ]]; then
     export CONTAINER_RUNTIME="docker"
     export BOOTSTRAP_CLUSTER="kind"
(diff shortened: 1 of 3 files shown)


Sunnatillo (commented):
/approve

fmuyassarov:
Did we change something that caused this issue ?

terror96:
> Did we change something that caused this issue ?

I couldn't tell: I can just see that a call to `run_clean.sh` in `jenkins/jobs/dynamic_fullstack_building.groovy` was added around the same time when the CI job started to fail.

EDIT: for me, it seems that all other scripts part of the same pipeline have the logic for a default value i.e. `OS_IMAGE` is not set in the pipeline env.

jenkins/scripts/dynamic_worker_workflow/run_clean.sh
Sunnatillo, line 21: We need to add  /tmp/dib/metal3-dev-env  location for fullstack test

terror96: Let's try to set `DEV_ENV_REPO_LOCATION` before building the fullstack.

adilGhaffarDev (commented):
/lgtm

PR #1346, opened 2026-05-29
Add curl retry and timeout opts

terror96:
Add configurable curl options shared across all Artifactory API calls:

- `CURL_CONNECT_TIMEOUT`: max time to establish connection
- `CURL_SPEED_LIMIT` / `CURL_SPEED_TIME`: abort on slow transfers
- `CURL_RETRY` / `CURL_RETRY_DELAY`: retry on transient errors
- `--continue-at -` to resume interrupted uploads

Refactor `rt_delete_artifact`, `rt_upload_artifact`, `rt_list_directory`, and `backup_old_image` to use the shared `CURL_COMMON_OPTS` array instead of repeating credentials and flags inline.

### DIFF
--- a/jenkins/image_building/upload-node-image.sh
+++ b/jenkins/image_building/upload-node-image.sh
@@ -2,20 +2,44 @@
 
 set -ex
 
+# How long curl waits for the initial connection/handshake to succeed (in seconds)
+CURL_CONNECT_TIMEOUT="${CURL_CONNECT_TIMEOUT:-60}"
+
+# Aborts the transfer if the speed drops below this limit (in bytes) for the specified duration (in seconds)
+CURL_SPEED_LIMIT="${CURL_SPEED_LIMIT:-104857600}"
+CURL_SPEED_TIME="${CURL_SPEED_TIME:-30}"
+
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

lentzi90:
/hold
Let's check how that node image upload goes. Setting hold since it is not a required test, so merging could happen before/without it passing.

tgerczei (commented):
/approve

smoshiur1237 (commented):
/lgtm

tuminoid:

Upload went fine.
/hold

terror96:
I think we still better implement @tgerczei's suggestion and use e.g. `numft` to convert `CURL_SPEED_LIMIT` from a human-readable format to a number of seconds in order to avoid confusion i.e. my intention was to put 1M and not 100M limit.

lentzi90:
Let's take it in as is and do a second PR with the numft addition

terror96:
> Let's take it in as is and do a second PR with the numft addition

Fix: https://github.com/metal3-io/project-infra/pull/1348

PR #1348, opened 2026-06-01
Add human-readable CURL_SPEED_LIMIT parsing

terror96:
Introduce parse_speed_limit() to convert IEC-suffixed values (e.g. 1M, 500K, 2G) to bytes before passing them to curl's --speed-limit option.

Uses numfmt --from=iec when available; falls back to manual arithmetic for K/M/G/T suffixes. Plain integers pass through unchanged. Invalid input falls back to the default of 1048576 (1 MiB).

### DIFF
--- a/jenkins/image_building/upload-node-image.sh
+++ b/jenkins/image_building/upload-node-image.sh
@@ -5,12 +5,13 @@ set -ex
 # How long curl waits for the initial connection/handshake to succeed (in seconds)
 CURL_CONNECT_TIMEOUT="${CURL_CONNECT_TIMEOUT:-60}"
 
-# Aborts the transfer if the speed drops below this limit (in bytes) for the specified duration (in seconds)
-CURL_SPEED_LIMIT="${CURL_SPEED_LIMIT:-104857600}"
+# Aborts the transfer if the speed drops below this limit (in bytes) for the specified duration (in seconds).
+# Default speed limit is 1 MiB/s and time is 30 seconds.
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

jenkins/image_building/upload-node-image.sh
```diff
@@ -26,26 +27,65 @@ CURL_COMMON_OPTS=(
...
+            return
+        fi
+    else
+        # Manual fallback: integer with K/M/G/T suffix (IEC binary)
+        if [[ "${value}" =~ ^([0-9]+)([KkMmGgTt])$ ]]; then
+            local num="${BASH_REMATCH[1]}"
+            local suffix="${BASH_REMATCH[2]}"
+            case "${suffix}" in
+                k|K) echo $(( num * 1024 )) ; return ;;
+                m|M) echo $(( num * 1048576 )) ; return ;;
```
tgerczei: I would prefer $(( num * 1024**2 )) here for improved readability - and similarly third and fourth powers of 1024 for gibibytes and tebibytes - but this is purely cosmetical.

tgerczei (commented):
There's a cosmetical nuance I've noted, but it's practically weightless. Nice work, kudos!

/approve

jenkins/image_building/upload-node-image.sh
terror96: This is actually a good suggestion. Let's do this change too before merging.

Sunnatillo (commented):
/lgtm

PR #1350, opened 2026-06-02
Prow: Upgrade CAPI to v1.11.11

lentzi90:
One minor version at a time. We keep CAPO on v0.12 (but take the latest patch), since v0.13 requires CAPI v1beta2 API to be available.


lentzi90:
These changes are applied automatically by Flux.

Sunnatillo (commented):
/approve

PR #1351, opened 2026-06-02
Pin Java 21 packages in CI images

terror96:
Add package locks/holds for JDK 21 on CentOS, openSUSE Leap, and Ubuntu CI images to prevent unintended upgrades during subsequent image builds or package updates.

- CentOS: Install dnf versionlock and pin java-21-openjdk
- Leap: Use zypper addlock to lock java-21-openjdk*
- Ubuntu: Use apt-mark hold to lock openjdk-21-jre


lentzi90 (commented):
/approve

Sunnatillo (commented):
/lgtm

Is this confirmed to be a cause of the issue?

lentzi90:
> Is this confirmed to be a cause of the issue?

Sort of. Metal3-dev-env updates all packages as part of every job that makes use of it. Since we have not pinned these packages, they will also get updated. Changing the binary in the middle of a running job can easily cause issues.

PR #1359, opened 2026-06-09
Continue only downloads

lentzi90:
For node image uploads and downloads, we added the option to continue if there was some issue. This makes sense to downloads, but can cause issues on upload. Especially when there is already an existing image that we want to overwrite.


lentzi90:
If I recall correctly, we do not upload node images in the node image building tests, so triggering it does nothing useful here. The normal test even less so.

terror96 (commented):
good, I was also thinking before if it is a problem but apparently yes. /lgtm

kashifest (commented):
/approve

adilGhaffarDev (commented):
/lgtm

terror96:
FYI: we can enable continue-at also for uploads, but it will require a slightly different approach. I'll be working on a new proposal a bit later.

PR #1362, opened 2026-06-11
Add 1.1.1.1 nameserver as fallback

### LINKED ISSUE #1284: Prow: Add second nameserver as fallback
We recently had an outage because the nameserver became unreachable. We should configure a second nameserver to avoid the SPOF.
This can be done here: https://github.com/metal3-io/project-infra/blob/75c646bd6037b112f3144ff34629e78230b549ba/prow/capo-cluster/openstackcluster.yaml#L41.
Note that we will need to bump the CAPO version first (https://github.com/metal3-io/project-infra/issues/1282), since this field is immutable in the current version. The latest v1.12 release includes the fix.
Another thing to note: This just configures the OpenStack network DHCP. It will not necessarily propagate to existing nodes. Therefore it would be good to verify after applying the change, and if necessary do a rollout of all machines.

UncivilizedGhost:
Add Cloudfare DNS nameserver in case Google's 8.8.8.8 server fails

Fixes: https://github.com/metal3-io/project-infra/issues/1284#issuecomment-4661140067

### DIFF
--- a/prow/capo-cluster/openstackcluster.yaml
+++ b/prow/capo-cluster/openstackcluster.yaml
@@ -39,6 +39,7 @@ spec:
   - cidr: 10.6.0.0/24
     dnsNameservers:
     - 8.8.8.8
+    - 1.1.1.1
   bastion:
     enabled: true
     spec:


tuminoid (commented):
/lgtm

Sunnatillo (commented):
/hold
/approve

/cc @lentzi90 PTAL before we merge

lentzi90:
Thanks, this looks good! Please keep the hold until I have time to apply the changes.

lentzi90 (commented):
Changes applied
/hold cancel

PR #1365, opened 2026-06-12
Use Artifactory copy API for image backups

terror96:
Optimize the image upload workflow to avoid downloading and re-uploading entire image files during backup and upload operations.

- Implement a server-side copy helper utilizing the Artifactory storage API
- Check file existence via the Artifactory API, eliminating the need to partially download files with curl to check their existence
- Upload images with a temporary name first, back up the prior image server-side, then safely copy/move the temporary file to its final destination
- Include a purging step to clean up any leftover temporary files from aborted or failed previous runs
- Use --continue-at with both upload and download operations


terror96:
/cc @lentzi90  @tgerczei

Submitting this as a draft because our artifactory does not currently support move/copy operations, but wanted to still `log` this for future cases.

lentzi90 (commented):
Probably we will have to make do without the move feature

Issue #1363, opened 2026-06-11
Leap basic e2e test on release-1.12 is failing


lentzi90:
Failing job: https://jenkins.nordix.org/view/Metal3%20periodic%20FAIL/job/metal3-periodic-leap-e2e-basic-test-release-1-12/

The test belongs to CAPM3 but we have reason to believe the issue is closer to image-building in this case. This is why I create the issue in this repo.

**Symptom:** Control plane Machines never appear. Serial logs stop at `Loading initial ramdisk`.

Initial finding and suspicion was that the node image was corrupted in Artifactory. We suspected that https://github.com/metal3-io/project-infra/pull/1346 could cause corruption and fixed that in https://github.com/metal3-io/project-infra/pull/1359.

We also went down the rabbit hole of cached images, suspected that the node image was stored inside the CI image and never downloaded.
This does not seem to be the case, but this is an area where we could be more transparent in the CI logs.

Currently the hypothesis is as follows.

## Summary of Root Cause Analysis

**The "Loading initramfs" stall is a red herring.** The node actually boots successfully. Here's what's really happening:

### 1. Serial Console Output is Missing (Not a Stall)

The serial console log shows:
```
Loading Linux 6.4.0-150600.23.115-default ...
Loading initial ramdisk ...
```
...and then nothing more. But this is because the **kernel command line lacks `console=ttyS0`**, so once the kernel takes over from GRUB, no output goes to the serial port. The node _is_ booting.

### 2. Proof the Node Boots Successfully

The `kubelet.log` and `cloud-final.log` from the artifacts prove the node booted:
```
Jun 11 01:07:04 test1-845qn cloud-init[1563]: Cloud-init v. 25.1.3-150400.15.7.6 running 'modules:final'
```
The node was up ~17 seconds before cloud-init's final stage ran.

### 3. The Real Failure: Missing Default Route

`kubeadm init` fails because there's no default route:
```
WARNING: could not obtain a bind address for the API Server: no default routes found in "/proc/net/route"
error: cannot use "0.0.0.0" as the bind address for the API Server
```

The IP pool (`baremetalv4-pool`) correctly specifies gateway `192.168.111.1` and prefix `24`, but the route is never applied to the node.

### 4. Root Cause: cloud-init Regression in Base Image

The base openSUSE Leap 15.6 cloud image was **freshly downloaded** during the node image build (job 89):
```
[...]

PR #1366, opened 2026-06-12
Add kernel parameters to print leap kernel log

nuhakala:
Add kernel command line parameters so that the kernel log is printed to serial log. The current leap issue (as well as any future issue) is hard to debug without the kernel log.

Related issue: #1363


kashifest:
/override metal3-node-image-building
lets first get more output from the test

kashifest (commented):
/approve

Rozzii (commented):
/lgtm

PR #1367, opened 2026-06-16
Disable wicked by default in leap node

nuhakala:
Disables wicked in the node image itself, similarly as it is disabled in the CI image.

Related to issue https://github.com/metal3-io/project-infra/issues/1363

I have investigated the failing e2e test and looks like the problem is faulty NetworkManager configurations. Cloud init does not render properly network manager configurations because network manager is not enabled. This is because wicked is used instead of network manager in the e2e tests. The dev env templates disable wicked and enable networkmanager in pre kubeadm commands.

Also, removes the DIB leap patching, because it is apparently using new enough version of DIB, which includes it already. DIB source code accepts versions 15.5 and 15.6. That patch is needed when building other versions but I think we don't need it ATM.


tuminoid (commented):
/lgtm
/cc @Rozzii

lentzi90 (commented):
/approve

PR #1369, opened 2026-06-18
Prow: Bump CAPI/CAPO versions

lentzi90:
Bumping CAPI to next minor version. Note that CAPO doesn't follow the same release schedule as CAPI. Bumping two minor versions for CAPO to "sync". CAPO v0.14 uses internally CAPI v1.12, so these are matching well.

Changes are applied automatically by flux.


kashifest (commented):
/approve

tuminoid (commented):
/lgtm

PR #1371, opened 2026-06-18
Prow: Bump Traefik to v3.7.5

lentzi90:
The biggest change here is that the Gateway API CRDs are no longer part of the chart.
Traefik still has its own CRDs though, so we need both.

Changes already applied.


kashifest (commented):
/approve

tuminoid (commented):
/lgtm

PR #1372, opened 2026-06-18
Fix renovate use glob syntax in managerFilePatterns instead of regex

kashifest:
fixes renovate since it is not opening PRs. Also removes the Monday schedule for now. Will put back once things are working  or if renovate creates too much traffic here.


tuminoid (commented):
/approve

lentzi90 (commented):
/lgtm

PR #1385, opened 2026-06-22
fix container image builds when building from release

tuminoid:
You can't overwrite GITHUB_ or RUNNER_ env vars, which we try to do which means when we release from branch, and do manual build + push from release branch, we end up pushing release branch content on top of main, because GITHUB_REF was expected to point to release branch but actually it keeps pointing to main, since it cannot be overwritten.

This leads to IRSO functional in ironic-image failing right now as there is microversion difference between main and 35.0 branches.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-environment-variable


lentzi90 (commented):
😮 For real!
/approve

smoshiur1237 (commented):
/lgtm

PR #1414, opened 2026-07-02
Modify container image build to allow PRs to be tested

UncivilizedGhost:
An image build workflow was added to multiple repositories based on this workflow in multiple repos:
- https://github.com/metal3-io/ip-address-manager/pull/1471
- https://github.com/metal3-io/ironic-ipa-downloader/pull/89
- https://github.com/metal3-io/ironic-standalone-operator/pull/733
- https://github.com/metal3-io/baremetal-operator/pull/3394

These workflows were failing as they didn't have secrets.

This Pr makea secrets optional

### DIFF
--- a/.github/workflows/container-image-build.yml
+++ b/.github/workflows/container-image-build.yml
@@ -63,11 +63,14 @@ on:
         default: false
     secrets:
       QUAY_USERNAME:
-        required: true
+        # Required for release/push jobs. PR build jobs that set pushImage: false can omit this.
+        required: false
       QUAY_PASSWORD:
-        required: true
+        # Required for release/push jobs. PR build jobs that set pushImage: false can omit this.
+        required: false
       SLACK_WEBHOOK:
-        required: true
+        # Required for release/push jobs to get failure notifications. PR build jobs can omit this.
+        required: false
     outputs:
       image-digest:
         description: "The digest of the built image"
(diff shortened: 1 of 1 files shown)


Sunnatillo:
Could you explain bit more in the description why we need this change?

.github/workflows/container-image-build.yml
peppi-lotta, line 187: Do we know how the 
```
username: ${{ secrets.QUAY_USERNAME }} # zizmor: ignore[secrets-outside-env]
password: ${{ secrets.QUAY_PASSWORD }} # zizmor: ignore[secrets-outside-env]
```
act when they are not set in the `Build ${{ inputs.image-name }} image` step? Does the build succeed? Should these values be omitted when they are not set?

.github/workflows/container-image-build.yml
```diff
@@ -63,11 +63,14 @@ on:
         default: false
     secrets:
       QUAY_USERNAME:
-        required: true
+        # Required for release/push jobs. PR build jobs that set pushImage: false can omit this.
```
peppi-lotta, line 66: This is a good premise but it's not guarded for in any way. Should we have some kind of check that these values are actually provided when `pushImage` is true or does the `mr-smithers-excellent/docker-build-push` handle that for us?

peppi-lotta (commented):
Thank you for opening the fix! I don't know how the `mr-smithers-excellent/docker-build-push` works exactly so I have some questions about how it reacts when the credentials are omitted. Maybe we need to have some more elaborate if else system for this.

.github/workflows/container-image-build.yml
UncivilizedGhost: I am not sure but I added an additional step which checks the secrets if `pushImage` is true. I think that should suffice.

smoshiur1237 (commented):
/lgtm

lentzi90 (commented):
/approve

Issue #1448, opened 2026-07-31
Consolidate CI orchestrators: move away from GitHub Actions


Sunnatillo:
We currently maintain three CI orchestrators: Jenkins, Prow, and GitHub Actions.

We started using GitHub Actions when we had a resource shortage in our CI. Now that CNCF has provided a good amount of resources in Oracle Cloud, we want to move away from GitHub Actions as much as possible and rely mainly on Jenkins and Prow.

The plan is to move lint checks, unit tests, and other smaller tests to Prow, and BMO e2e-like tests to Jenkins.

**Why**:

- For larger tests, Jenkins is the better fit: the pipeline is clear and the logs are easier to read.
- Prow acts as a gating tool (lgtm/approve).
- Running small tests in two orchestrators is redundant.

**Approach**: add the Prow test first, then remove the GitHub Actions one, step by step.

- [ ]  Unit tests → Prow
- [ ]  BMO e2e tests → Jenkins
- [ ]  PR verifier → Prow
- [ ]  golangci-lint → Prow
- [ ]  yamllint → Prow

 ...

Sunnatillo:
https://github.com/metal3-io/project-infra/pull/1453
https://github.com/metal3-io/baremetal-operator/pull/3481
https://github.com/metal3-io/baremetal-operator/pull/3479

Ahmedaltu:
Ticket Update:

- golangci-lint — added ✅, enforced, GHA removed
- pr-verifier — added ✅, enforced, GHA  removed
- unit — added ✅, enforced, GHA removed
- yamllint — added partially, not yet enforced,  GHA still in place

lentzi90:
Reverting to GH actions until we have a solution for pr-verifier and batch jobs.
When multiple PRs are about to be merged at the same time, prow will do them in batches. It then merges together all PRs in a batch and runs the tests on the resulting merge commit. In this case there is no PULL_TITLE parameter set and so pr-verifier fails.

lentzi90:
Working around the issue in https://github.com/metal3-io/project-infra/pull/1496

Sunnatillo:
We have faced an issue with PR verifier job, when merging PR prow runs jobs in batch. Test cannot retrieve PR title from batch jobs. This issue was blocking the merges. 
We have add condition not to run PR verifier for batch jobs. It solved the issue.

https://github.com/metal3-io/project-infra/pull/1496

Ahmedaltu:
[...]

PR #1453, opened 2026-08-03
Add golangci-lint presubmit for baremetal-operator

Ahmedaltu:
Part of #1448. 
Adds a Prow golangci-lint presubmit for BMO, added alongside the existing GitHub Actions lint. Once it's green the GHA workflow will be removed.


lentzi90 (commented):
I would suggest to make this optional while we verify that it is working. I.e. set `optional: true` for it. Then once we see that it passes we can enforce it.

Ahmedaltu:
updated..

lentzi90 (commented):
/approve

Sunnatillo (commented):
/lgtm

PR #1455, opened 2026-08-03
make version check workflow work with other repos

Rozzii:
This commit:
 - Makes sure that ./hack/versioncheck.sh is present during the container-image-build execution even if it is called as a reusable workflow from an other repository.
 - Build action also seems to be buggy and does not handle the more than
   one label correctly so the revision will be removed for the time being
   until I sort out the issue

Reason:
 - Workflow breaks without the change when called from other repo's workflow pipeline.


Rozzii:
/approve
There are no approver online other than me, and image builds are failing, in order to unblock the CI I have to approve my own PR.

PR #1456, opened 2026-08-04
Add golangci-lint presubmit for ironic-standalone-operator

Ahmedaltu:
Part of #1448. 
Adds a Prow golangci-lint presubmit for ironic-standalone-operator, optional for now. Once it's green the GitHub Actions workflow will be removed.

PR #1460, opened 2026-08-05
Add golangci-lint presubmit for ironic-standalone-operator release branches

Ahmedaltu:
Part of #1448, follow-up to #1456. Adds the golangci-lint presubmit to ironic-standalone-operator release-0.9 and release-0.10, matching the main-branch job in #1456.


lentzi90 (commented):
/approve
I have not checked if we used to have go 1.25 on these branches, but since this is optional for now, we might as well just check if it works.

Ahmedaltu:
The go image matches what the other jobs in these release files already use, so it should be consistent with the branch.
And Prow run will confirm it either way.

Sunnatillo (commented):
/lgtm

PR #1465, opened 2026-08-07
Enforce golangci-lint presubmit for ironic-standalone-operator

Ahmedaltu:
Follow-up to #1456. golangci-lint has been running green on IrSO PRs, so this makes it required (removes optional: true). Pairs with the GitHub Actions golangci removal so lint stays enforced via Prow.

Part of #1448.


Sunnatillo (commented):
/approve

lentzi90 (commented):
/lgtm

PR #1467, opened 2026-08-10
Fix broken container-image-build workflow

UncivilizedGhost:
GitHub Actions does not allow secrets.* in if: expressions. Use inputs.pushImage instead.

### DIFF
--- a/.github/workflows/container-image-build.yml
+++ b/.github/workflows/container-image-build.yml
@@ -264,7 +264,7 @@ jobs:
         path: ${{ inputs.image-name }}-sbom.spdx.json
 
     - name: Slack Notification on Failure
-      if: ${{ failure() && secrets.SLACK_WEBHOOK != '' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/container-image-build.yml
```diff
@@ -264,7 +264,7 @@ jobs:
         path: ${{ inputs.image-name }}-sbom.spdx.json
 
     - name: Slack Notification on Failure
-      if: ${{ failure() && secrets.SLACK_WEBHOOK != '' }}
+      if: ${{ failure() && inputs.pushImage }}
```
tuminoid, line 267: Is the intent checking for any value for push image or true'ish?

UncivilizedGhost: push image is a boolean so it's just checking if it's true

smoshiur1237 (commented):
/lgtm

Rozzii (commented):
/approve
Should be reasonable change

PR #1491, opened 2026-08-18
Add Loki push credential to BMO e2e job

migiFi:
Binds the metal3-ci-log-collector-push credential in the BMO optional e2e job so ci-2e2.sh can stream e2e logs to the CI log collector (Loki).

Has no effect on its own, log streaming only starts once the BMO change is also merged.

### DIFF
--- a/jenkins/jobs/bmo_e2e_optional_tests.groovy
+++ b/jenkins/jobs/bmo_e2e_optional_tests.groovy
@@ -9,6 +9,7 @@ def refspec = '+refs/heads/' + pullBase + ':refs/remotes/origin/' + pullBase + '
 pipeline {
     environment {
         GINKGO_FOCUS = "${GINKGO_FOCUS}"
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
Do we need to set LOKI_ADDR also? I mean it is not a credential, but it makes sense to set that in the same place I think.

migiFi:
> Do we need to set LOKI_ADDR also?

Added `LOKI_URL` now, thats the variable the Alloy config reads for the push endpoint (`LOKI_ADDR` is logcli's). [run-alloy.sh](https://github.com/metal3-io/baremetal-operator/pull/3538/changes#diff-14b66517caaaf972cec0d3957a60d6ce05e1876a0306f356b55b533bf23ec620) has a staging default for the local runs, so the pipeline value is preferred in CI.

lentzi90 (commented):
/approve

kashifest (commented):
/lgtm

PR #1496, opened 2026-08-20
prow: do not run pr verifier in batch jobs

Sunnatillo:
In batch jobs verifier test cannot retrieve PR title and fails the PR verifier test. Since it is already testing in individual PR level, we do not necessarily need to test  it before merge.


lentzi90 (commented):
/approve

adilGhaffarDev (commented):
/lgtm

PR #1499, opened 2026-08-20
Auto-trigger BMO e2e presubmit (keep optional)

Ahmedaltu:
Set always_run: true on metal3-bmo-e2e-test-optional-pull so the Jenkins-backed BMO e2e job runs on every PR. 
Keeps optional: true — this run alongside the existing GitHub Actions e2e-test, not a merge gate yet. 

Part of #1448 (move BMO e2e off GitHub Actions).

### DIFF
--- a/prow/config/jobs/metal3-io/baremetal-operator.yaml
+++ b/prow/config/jobs/metal3-io/baremetal-operator.yaml
@@ -284,7 +284,7 @@ presubmits:
     branches:
     - main
     agent: jenkins
-    always_run: false
+    always_run: true
     optional: true
   # name: {job_prefix}-{image_os}-e2e-integration-k8s-pre-release-test-{capm3_target_branch}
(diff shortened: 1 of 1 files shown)


prow/config/jobs/metal3-io/baremetal-operator.yaml
Sunnatillo, line 283: These are optional test and should not run every time.

Sunnatillo (commented):
/hold

prow/config/jobs/metal3-io/baremetal-operator.yaml
Ahmedaltu: @Sunnatillo In order to move e2e tests from GHA to Jenkins, and since the Jenkins e2e groovy job already covers what the GHA e2e job does, we just need to make it always_run so it runs in parallel with the existing GHA job. Still optional:true, so it's not gating merges yet, just building up a track record before we can drop the GHA workflow.

PR #1507, opened 2026-08-27
Make resource cleanup run in OCI

tgerczei:
Historically these jobs were run in Xerces, against Xerces. They have recently been made to run against OCI, but still run in Xerces. Adjust labels so that they will run in  OCI.


lentzi90 (commented):
/approve

adilGhaffarDev (commented):
/lgtm

PR #1510, opened 2026-08-28
Split CAPM3 lint into two Prow jobs

migiFi:
Update the CAPM3 presubmits so linting runs as two independent checks. The golangci-lint job now runs `make golangci` (golangci-lint across all modules no KAL) instead of `make lint`, and a new `kube-api-lint` job runs make `make kube-api-lint` (KAL only).

This avoids running KAL twice and keeps the two checks separate.

**NOTE**: This is PR(2) depends on PR(1) [#3656](https://github.com/metal3-io/cluster-api-provider-metal3/pull/3656) make targets existing on CAPM3 main. These jobs will fail until PR(1) is merge.

### DIFF
--- a/prow/config/jobs/metal3-io/cluster-api-provider-metal3.yaml
+++ b/prow/config/jobs/metal3-io/cluster-api-provider-metal3.yaml
@@ -121,7 +121,24 @@ presubmits:
     spec:
       containers:
       - args:
-        - lint
+        - golangci
+        command:
+        - make
+        env:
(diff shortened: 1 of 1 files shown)


Sunnatillo (commented):
@migiFi We need to create the other test as well on the same PR, otherwise we will not test PRs with golangci lint tests

Sunnatillo (commented):
/approve

adilGhaffarDev (commented):
/lgtm

PR #1513, opened 2026-08-29
Add BMO e2e fixture test Prow presubmit (Jenkins)

Ahmedaltu:
Adds `metal3-bmo-e2e-fixture-test-optional-pull`, a new Jenkins-backed
Prow presubmit for baremetal-operator that runs the e2e fixture suite
(`make docker && make test-e2e` against `test/e2e/config/fixture.yaml`),
mirroring what `.github/workflows/e2e-fixture-test.yml` already does
on GitHub Actions.

Part of #1448 (move BMO e2e off GitHub Actions onto Jenkins/OCI).

### DIFF
--- a/jenkins/jobs/bmo_e2e_fixture_test.groovy
+++ b/jenkins/jobs/bmo_e2e_fixture_test.groovy
@@ -1,0 +1,48 @@
+// Set defaults for non-PR jobs
+def pullSha = (env.PULL_PULL_SHA) ?: 'main'
+def pullBase = (env.PULL_BASE_REF) ?: 'main'
+def repoUrl = 'https://github.com/metal3-io/baremetal-operator.git'
+// Fetch the base branch and the pullSha, nothing else
+def refspec = '+refs/heads/' + pullBase + ':refs/remotes/origin/' + pullBase + ' ' + pullSha
+
+pipeline {
+    agent { label 'metal3ci-8c32gb-ubuntu-oci' }
+    environment {
+        E2E_CONF_FILE = "${WORKSPACE}/test/e2e/config/fixture.yaml"
+        GINKGO_NODES = '1'
+        IMG = 'quay.io/metal3-io/baremetal-operator'
+        IMG_TAG = 'e2e'
+    }
+    stages {
+        stage('Checkout source code') {
+            steps {
+                deleteDir()
+                checkout scmGit(
+                    branches: [[name: pullSha]],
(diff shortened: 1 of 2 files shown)


prow/config/jobs/metal3-io/baremetal-operator.yaml
```diff
@@ -288,6 +288,15 @@ presubmits:
       jenkins_operator: oci
     always_run: false
     optional: true
+  - name: metal3-bmo-e2e-fixture-test-optional-pull
+    branches:
+    - main
+    agent: jenkins
+    labels:
+      jenkins_operator: oci
+    run_if_changed: '(test/e2e/.*|pkg/.*|internal/.*|cmd/.*|config/overlays/fixture/.*|Makefile|go\.(mod|sum))$'
```
lentzi90, line 297: I think we are missing API changes here. Maybe also some scripts that could be called from the Makefile. Perhaps a safer alternative would be to skip only if markdown files changed? These are quite cheap tests that we usually want to run (fixture are not as heavy as the "full" e2e).

jenkins/jobs/bmo_e2e_fixture_test.groovy
```diff
@@ -0,0 +1,48 @@
...
+    }
+    stages {
+        stage('Checkout source code') {
+            steps {
+                deleteDir()
+                checkout scmGit(
+                    branches: [[name: pullSha]],
+                    userRemoteConfigs: [[url: repoUrl, refspec: refspec]],
+                    extensions: [[$class: 'CleanCheckout'],
+                    [$class: 'CleanBeforeCheckout'],
```
lentzi90, line 24: Please coordinate this with @Sunnatillo so we don't end up adding more deprecated things at the same time that we remove it in https://github.com/metal3-io/project-infra/pull/1516

prow/config/jobs/metal3-io/baremetal-operator.yaml
Ahmedaltu: You are right, i will switch this to skip-if-only-markdown so it runs broadly.

jenkins/jobs/bmo_e2e_fixture_test.groovy
Ahmedaltu: I'll hold this PR until #1516 merges, then apply the deleteDir pattern.

PR #1524, opened 2026-09-03
add GH token to SCM steps in pipeline scripts

Rozzii:
This PR:

 - Introduces the use of an existing GitHub token for the declarative SCM steps in the pipeline scripts.

This change is needed to avoid hitting GitHub rate limit.

### DIFF
--- a/jenkins/jobs/bml_integration_tests.groovy
+++ b/jenkins/jobs/bml_integration_tests.groovy
@@ -52,7 +52,7 @@ pipeline {
                         [$class: 'CleanBeforeCheckout']
                         ],
                     submoduleCfg: [],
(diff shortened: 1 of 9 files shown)


Rozzii:
This test runs into the problem that this PR tries to solve but the CI does not inject the pipeline scripts from this PR but from main.
So the test is not relevant-

Sunnatillo (commented):
/approve

Sunnatillo:
/hold
I we need to fix all issues in one PR.

Rozzii:
> /hold I we need to fix all issues in one PR.

No need to fix everything in 1 PR. This specifically just fixes issues with the SCM step, other issues might not be in project infra.

lentzi90 (commented):
/lgtm

PR #1532, opened 2026-09-07
Add Slack receiver for Alertmanager on Default and Critical

### LINKED ISSUE #930: Prow: Alerting
Split out from #896 

When #896 is done, we will have the basics in place that will allow us to define alerts.
This issue is for doing that.
Define some basic alerts (or use those that comes with kube-prometheus), set up [receivers](https://prometheus.io/docs/alerting/latest/configuration/#receiver-integration-settings) for slack and/or email.
Useful alerts would be for example

- generic kubernetes issues like unhealthy nodes (see [this](https://github.com/kubernetes/k8s.io/blob/98668f8f2003a19fb9fbd6d5bc1e4555a30d4086/infra/aws/terraform/prow-build-cluster/resources/monitoring/alertmanager-main/prometheus-rule-node-status.yaml))
- unhealthy workloads (except tests, they can be expected to fail), for example, prow,

Ahmedaltu:
Fixes #930

Wires up an actual notification destination for Prow's Alertmanager alerting stack, which has been deployed since #896 but never had anywhere to send alerts.

### DIFF
--- a/prow/bootstrap/clustersecretstore.yaml
+++ b/prow/bootstrap/clustersecretstore.yaml
@@ -18,3 +18,4 @@ spec:
     - kube-system # CPO runs here
     - prow # Prow components run here
     - test-pods # Pods for ProwJobs run here
+    - monitoring # Alertmanager Slack webhook secret
--- a/prow/infra/kube-prometheus/README.md
+++ b/prow/infra/kube-prometheus/README.md
@@ -10,6 +10,7 @@ This is how you apply it in the cluster:
 ```bash
 kubectl apply -f manifests/setup
 kubectl apply -f manifests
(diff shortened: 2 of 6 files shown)


lentzi90 (commented):
Can we do this after the migration to OCI?
The monitoring stack has been very neglected. I am not sure it is working at the moment and it would be better to first have prow on community controlled infra. Then alerts could also be handled by the community.

Ahmedaltu:
@lentzi90  yes, we could leave to after OCI migration. should it be closed or leave it open?

lentzi90:
> @lentzi90 yes, we could leave to after OCI migration. should it be closed or leave it open?

It doesn't matter to me if we close or leave it, but it will probably require some changes after the migration, so in that sense perhaps better to just close it.

PR #1549, opened 2026-09-15
🐛 leap-node: use NetworkManager renderer for cloud-init

adilGhaffarDev:
The leap-node image switches from wicked to NetworkManager, but cloud-init still defaults to the sysconfig (wicked ifcfg) renderer, which NM does not read. Config-drive network settings (static IP + default route) are written but never applied, so the node DHCPs without a default route and kubeadm fails to bind the API server.

What this PR is doing:
Pin cloud-init to the network-manager renderer/activator so it writes NM keyfiles that NetworkManager applies.


adilGhaffarDev:
seems like job is not setup properly: https://jenkins.apps.test.metal3.io/job/metal3-node-image-building/

Sunnatillo:

Not testing the changes

Sunnatillo (commented):
/approve

kashifest (commented):
/lgtm

PR #1556, opened 2026-09-18
Enable crio service in node image builds

### LINKED ISSUE #1554: Enable crio in node images
We should enable crio in the node images. We already install it, but now users have to enable/start it themselves.
There is also a discrepancy between centos and ubuntu here. The ubuntu package automatically enables crio, while the centos package does not.
We should just make sure it is always enabled so we can avoid extra things like this:
https://github.com/metal3-io/cluster-api-provider-metal3/blob/8d353f0097b5f1e4352bdb734fb9457c2fd0966c/test/e2e/data/infrastructure-metal3/main/bases/centos-kubeadm-config/centos-kubeadm-config.yaml#L143
in other places.
We can do it immediately after installing here: https://github.com/metal3-io/project-infra/blob/918ef7afa3821154b4e1649918f6482641aed54b/jenkins/image_building/dib_elements/centos-node/install.d/55-install#L51

Ahmedaltu:
Enable crio service in node image builds.
Fixes #1554


lentzi90 (commented):
This looks good!
/approve
However, considering the CI situation at the moment, I don't think we can merge this until we figure out a new way of automating node image builds and uploads
/hold

Sunnatillo (commented):
/lgtm

smoshiur1237 (commented):
/lgtm

